PR #9, opened 2018-12-14
Add fullnameOverride option

tetafro:
I've modified `fullname` template to add `fullnameOverride`. It makes possible to use short names like just `elasticsearch` service instead of `my-release-elasticsearch`.

The code is taken from template generated by `helm create` command.


tetafro:
Sorry, didn't notice that there is `uname` template, which is as a name for resources.

PR #17, opened 2018-12-21
Add https support to readinessProbe

### LINKED ISSUE #16: Kibana readinessProbe fails when server.ssl.enabled=true
The [readinessProbe](https://github.com/elastic/helm-charts/blob/master/kibana/templates/deployment.yaml#L55-L73) only appears to target HTTP, not HTTPS.

jordansissel:
If server.ssl.enabled is set in Kibana (vi SERVER_SSL_ENABLED env var),
then the readinessProbe will use https to test for readiness.

Fixes #16

### DIFF
--- a/kibana/values.yaml
+++ b/kibana/values.yaml
@@ -53,6 +53,10 @@ service:
   type: ClusterIP
   port: 5601
 
+# Set this if you are setting server.ssl.enabled in Kibana.
+# This value is a hostname accepted by the SSL certificate provided to kibana.
+kibanaSSLHostname: localhost
+
 ingress:
   enabled: false
   annotations: {}
--- a/kibana/values.yaml
+++ b/kibana/values.yaml
@@ -55,6 +55,7 @@ service:
 
 # Set this if you are setting server.ssl.enabled in Kibana.
 # This value is a hostname accepted by the SSL certificate provided to kibana.
+# Configuring this allows the readinessProbe to successfully check Kibana over HTTPS.
 kibanaSSLHostname: localhost
 
 ingress:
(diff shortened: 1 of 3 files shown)


jordansissel:
Caveat: This is my first effort to do anything of substance with Helm.

Open questions:
* Is `kibanaSSLHostname` the best value name?
* Does the shell script need more comments?

jordansissel:
Test results:

```
% kubectl get pods -l app=kibana
NAME                            READY   STATUS    RESTARTS   AGE
xyz-kibana-5dfb9fcdb8-pfv6l   1/1     Running   0          15m
xyz-kibana-5dfb9fcdb8-s6zsq   1/1     Running   0          15m
```

(Without this patch, the `READY` state would forever be `0/1` as the check over http would always fail)

kibana/values.yaml
```diff
@@ -53,6 +53,10 @@ service:
   type: ClusterIP
   port: 5601
 
+# Set this if you are setting server.ssl.enabled in Kibana.
+# This value is a hostname accepted by the SSL certificate provided to kibana.
+kibanaSSLHostname: localhost
```
Crazybus: Could you also this to the [configuration](https://github.com/elastic/helm-charts/tree/master/kibana#configuration) section of the readme?

Crazybus:
> Is kibanaSSLHostname the best value name?

Sounds right to me and it is following the helm variable naming conventions. 

> Does the shell script need more comments?

The code is clear and logical to me, I didn't have any trouble understanding it without comments 👍

kibana/values.yaml
jordansissel: +1 will do.

Crazybus:
Thanks for adding your config! I'll update the integration test/example to use it.

PR #21, opened 2018-12-28
Add testing suite for v7 using the new zen2 discovery method

Crazybus:


elasticsearch/README.md
```diff
@@ -43,6 +43,7 @@ This helm chart is a lightweight way to configure and run our official [Elastics
...
 | `minimumMasterNodes`      | The value for [discovery.zen.minimum_master_nodes](https://www.elastic.co/guide/en/elasticsearch/reference/current/discovery-settings.html#minimum_master_nodes). Should be set to `(replicas / 2) + 1`                                                                                                                    | `2`                                                                                                                       |
+| `zen2Discovery`           | Enabling this will set `cluster.initial_master_nodes` to a list of all of the masters. Only needed for Elasticsearch >= 7 | `false`                                                                                                                       |
```
DaveCTurner: We would like to keep the name `zen2` out of the UI as much as possible, intending it to be the default (and only) cluster coordination mechanism in future versions. I think this behaviour should be enabled if-and-only-if the Elasticsearch version is ≥7. Can we make this a function of the version (e.g. by looking at the `imageTag` property)?

Crazybus: Because we can't guarantee what `imageTag` users will be using for custom images that won't be possible. Anyone wanting to install plugins into a docker image has to create their own custom images. I think the time has come for a dedicated `esVersion` variable (`6.x`, `7.x`). I had been avoiding it till now because there wasn't any version specific differences in the helm chart.

Crazybus: Fixed! Thanks for catching that.

PR #22, opened 2018-12-29
Rewrite elasticsearch multi example to use child charts.

jordansissel:
As a newbie to helm, I found child charts interesting. This change makes the `multi` example now use a single release with child charts instead of two separate helm releases.

In practice, I don't know if this is the best route, because we'll want to upgrade masters before upgrading data nodes, for example, so this PR is mostly food-for-thought.

Thoughts? :)


jordansissel:
> we'll want to upgrade masters before upgrading data nodes

I suppose in this example, we could set `imageTag` separately for each child chart in the values.yaml and do an upgrade that way:

* set imageTag on master to new version
* `make install` to upgrade just the masters
* set imageTag on data role to new version
* `make install` to ugprade just the data nodes

It'd be identical to the the upgrade process for current code (in master) which has two files (data + master) and separate helm releases, probably?

jordansissel:
> If you also wanted to add Kibana to this setup

Indeed! I'll share my stack chart with you after new years which includes Kibana.

elasticsearch/examples/multi/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+description: Example elasticsearch with separate data and master nodes implemented with child charts.
+home: https://github.com/elastic/helm-charts
+maintainers:
```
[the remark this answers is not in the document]
jordansissel: Done!

jordansissel:
Rebased!

Crazybus:
CI tests are [failing](https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+integration-elasticsearch/13/ES_SUITE=multi,KUBERNETES_VERSION=1.11,label=docker&&virtual/console) because helm isn't initialised. 

```
22:54:45 make[1]: Entering directory `/app/elasticsearch/examples/multi'
22:54:45 helm dep update
22:54:45 Error: Couldn't load repositories file (/app/.helm/repository/repositories.yaml).
22:54:45 You might need to run `helm init` (or `helm init --client-only` if tiller is already installed)
```

I think adding the `helm init --client-only` into the testing `Dockerfile` would be the cleanest way to fix it. That is defined in [helpers/terraform/Dockerfile](https://github.com/elastic/helm-charts/blob/bf17d9bf0d4b48e41e50a098f75c37a4b7e4edac/helpers/terraform/Dockerfile#L38)

rendhalver:
Doesn't this go against the original design of the Chart?
One of the original intents for having multiple releases was to be able to swap out the data node without needing to delete the data. If we have one release for a cluster we could have the same problem.
https://github.com/elastic/helm-charts/tree/master/elasticsearch#usage-notes-and-getting-started

Crazybus:
> Doesn't this go against the original design of the Chart?
One of the original intents for having multiple releases was to be able to swap out the data node without needing to delete the data. If we have one release for a cluster we could have the same problem.

This still functions identically to the previous example. Instead of adding a new release and a new upgrade command you can instead add a new node block. 

```
es-data2:
  <<: *defaults # include with the default settings
  nodeGroup: "data2"

  roles:
    master: "false"
    ingest: "true"
    data: "true"
```

To then demote the old nodes you would then bump the replicas down one at a time, or manually move the shards off them then remove the node block. 

While this is the same in function the terminology in the readme indeed no longer matches. I think we can replace the terminology "release" with "install". The important piece being that the chart itself does not attempt to handle multiple groups of node itself. 

Upgrading the master group first would involve bumping the `imageTag` for only the masters first and still works in the same way as it would have before. 

The motivation for doing it this way was to remove code duplication when specifying configuration that is shared between multiple node groups. This could also be solved by including multiple values files.

One advantage of using this method is that it gives you a single version for your cluster. If you want to roll back the previous change there is only one version to rollback. Rather than manually needing to figure out the order that you made the changes to the separate releases. 

@rendhalver If you do find this example confusing though maybe it is a sign that we shouldn't merge this. Or maybe instead have it as a separate example of how to manage multiple node groups in a single release.

rendhalver:
It's not confusing I understand what you are trying to do.

I do admit I didn't read what the code does before I put in my comment.

I think having it as a separate example is a better idea.

If we start enforcing opinions then we are at risk of breaking the un-opinionated nature of the Chart.

rendhalver:
On a side note I copied this Chart into our local repo and we have used it to deploy 4 new clusters and it works really well.
I vote for adding this as a new top level chart so people can make use of it.

jordansissel:
@rendhalver I have some other ideas to make this chart style even more useful (so that settings can be inherited across parent->child chart values). Hopefully will have a PR to demo it this week. After that, we can discuss any future support for this style.

And thank you for testing! :)

rendhalver:
That sounds pretty cool!
Tag me on your new PR. :)

jordansissel:
@Crazybus I think this is ready for another round of reviews. CI tests failure seems unrelated ([timeout waiting for the GKE cluster to come online](https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+cluster-creation/72/KUBERNETES_VERSION=1.11,label=docker&&virtual/console))

desaintmartin:
Any update? Do you need help on this one?

elasticsearch/examples/multi/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  roles:
+    master: "true"
+    ingest: "false"
+    data: "false"
+
+# To define common settings, it may be useful to use YAML anchors to use the same 
+# base settings for both node types (data + master)
+es-defaults: &defaults
+  clusterName: "multi"
+  masterService: "multi-master"
```
desaintmartin: Since `es-master.nodeGroup` is master, do we need this one?

elasticsearch/examples/multi/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+es-master:
+  <<: *defaults # include with the default settings
+  nodeGroup: "master"
+
+  roles:
+    master: "true"
+    ingest: "false"
+    data: "false"
+
+# To define common settings, it may be useful to use YAML anchors to use the same
```
desaintmartin: In helm 2.14.3, the anchor has to be defined BEFORE using it, so moving this part on top of the file works.

Crazybus:
> Any update? Do you need help on this one? I can help bringing a multi example with several node groups (hot/warm/cold)

Pleas do! (In a separate PR).

> I think having it as a separate example is a better idea.

I agree with @rendhalver and think that we should add this is a example to the existing multi one that is being used for testing.

desaintmartin:
> > Any update? Do you need help on this one? I can help bringing a multi example with several node groups (hot/warm/cold)
> 
> Pleas do! (In a separate PR).

5 months later, PR at https://github.com/elastic/helm-charts/pull/22#issuecomment-532657384

PR #24, opened 2018-12-31
Switch to goss for integration testing of the Elasticsearch examples

### LINKED ISSUE #11: helm test fails when security is enabled
```
% helm ls elasticsearch-data
NAME                    REVISION        UPDATED                         STATUS          CHART                           APP VERSION     NAMESPACE
elasticsearch-data      3               Sat Dec 15 21:30:24 2018        DEPLOYED        elasticsearch-6.5.3-alpha1      6.5.3           default  
```

Running helm test:
```
% helm test elasticsearch-data
RUNNING: elasticsearch-data-rwrbf-test
FAILED: elasticsearch-data-rwrbf-test, run `kubectl logs elasticsearch-data-rwrbf-test --namespace default` for more info
Error: 1 test(s) failed           
```

Helm test logs:
```
% kubectl logs elasticsearch-data-rwrbf-test --namespace default
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (52) Empty reply from server                  
```

The curl command doesn't use https (nor configured w/ credentials and certificates)
```
% kubectl get pod elasticsearch-data-rwrbf-test -o yaml | grep curl
      curl -XGET --fail 'redacted:9200/_cluster/health?wait_for_status=green&timeout=1s'
```

https enabled in the pods
```
% kubectl get pod redacted-0 -o yaml | grep -A1 security.http.ssl.enabled
    - name: xpack.security.http.ssl.enabled
      value: "true"
```

Crazybus:
Closes #11 

Only thing I'm not sure about is whether or not to remove the current `helm test` pod. For the integration testing it doesn't add any value. But for cluster operators it does give them a very generic easy way to run `/_cluster/health`. Thoughts?

### DIFF
--- a/helpers/examples.mk
+++ b/helpers/examples.mk
@@ -1,0 +1,6 @@
+GOSS_VERSION := v0.3.6
+
+goss:
+	kubectl cp test/*.yaml $(GOSS_CONTAINER):/tmp/goss.yaml
+	kubectl exec $(GOSS_CONTAINER) -- sh -c "cd /tmp/ && curl -s -L https://github.com/aelsabbahy/goss/releases/download/$(GOSS_VERSION)/goss-linux-amd64 -o /usr/local/bin/goss && chmod +rx /usr/local/bin/goss && goss validate --color --format documentation"
+
--- a/helpers/examples.mk
+++ b/helpers/examples.mk
@@ -2,5 +2,5 @@ GOSS_VERSION := v0.3.6
 
 goss:
 	kubectl cp test/*.yaml $(GOSS_CONTAINER):/tmp/goss.yaml
-	kubectl exec $(GOSS_CONTAINER) -- sh -c "cd /tmp/ && curl -s -L https://github.com/aelsabbahy/goss/releases/download/$(GOSS_VERSION)/goss-linux-amd64 -o /usr/local/bin/goss && chmod +rx /usr/local/bin/goss && goss validate --color --format documentation"
+	kubectl exec $(GOSS_CONTAINER) -- sh -c "cd /tmp/ && curl -s -L https://github.com/aelsabbahy/goss/releases/download/$(GOSS_VERSION)/goss-linux-amd64 -o goss && chmod +rx ./goss && ./goss validate --color --format documentation"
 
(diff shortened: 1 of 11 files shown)


helpers/examples.mk
```diff
@@ -0,0 +1,6 @@
+GOSS_VERSION := v0.3.6
+
+goss:
+	kubectl cp test/*.yaml $(GOSS_CONTAINER):/tmp/goss.yaml
+	kubectl exec $(GOSS_CONTAINER) -- sh -c "cd /tmp/ && curl -s -L https://github.com/aelsabbahy/goss/releases/download/$(GOSS_VERSION)/goss-linux-amd64 -o goss && chmod +rx ./goss && ./goss validate --color --format documentation"
```
jordansissel: While fiddling with this, I had an idea to use `kubectl run` for testing the service if we need such a thing:

```
IMAGE := docker.elastic.co/elasticsearch/elasticsearch:6.5.4

goss:
	tar -C test -zc . | kubectl run -i --attach --generator=run-pod/v1 --rm --image $(IMAGE)  helm-test -- sh -c 'cd $(mktemp -d); curl -vOL https://github.com/aelsabbahy/goss/releases/download/v0.3.6/goss-linux-amd64; chmod +x ./goss-linux-amd64; tar -zxv;  ./goss-linux-amd64 validate --color --format documentation'
```

This copies the files using a tar pipe over stdin through kubectl. Some changes would need to be made, though, since everything is `localhost` and you would want to refer to `elasticsearch-master` as the hostname. The tests fail largely due to `localhost` and kernel/filesystem checks not passing since it's not running in the elasticsearch-master container.

jordansissel:
Tested manually on the `default` example:

```
% make goss
...
Total Duration: 0.008s
Count: 16, Failed: 0, Skipped: 0
```

Cool :)

helpers/examples.mk
Crazybus: > While fiddling with this, I had an idea to use kubectl run for testing the service if we need such a thing

We could also test the service from within the same container too right? As in adding another `http` test which hits the service instead of localhost. Testing that the service works from another container wouldn't really add anything unless I'm missing something.

Crazybus:
> With regards to the test pod, I'm kind of inclined to leave it in with a note somewhere in the README about it.

Sounds good to me. I have made a note to add this in a followup PR.

PR #26, opened 2018-12-31
Add https to the security example for Kibana

Crazybus:
Implements the config from https://github.com/elastic/helm-charts/pull/17#issuecomment-450467054

### DIFF
--- a/kibana/examples/security/Makefile
+++ b/kibana/examples/security/Makefile
@@ -3,8 +3,16 @@ default: install
 
 RELEASE := helm-kibana-security
 
-install:
+install: secrets
 	helm upgrade --wait --timeout=600 --install --values ./security.yml $(RELEASE) ../../ ; \
 	
 purge:
 	helm del --purge $(RELEASE)
+
+secrets:
+	kubectl delete secret kibana-certificates || true
+	vault read -field=kibana.crt secret/devops-ci/helm-charts/kibana/security/certificates | base64 --decode > kibana.crt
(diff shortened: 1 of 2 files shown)


kibana/examples/security/Makefile
```diff
@@ -3,8 +3,16 @@ default: install
...
-install:
+install: secrets
 	helm upgrade --wait --timeout=600 --install --values ./security.yml $(RELEASE) ../../ ; \
 	
 purge:
 	helm del --purge $(RELEASE)
+
+secrets:
+	kubectl delete secret kibana-certificates || true
+	vault read -field=kibana.crt secret/devops-ci/helm-charts/kibana/security/certificates | base64 --decode > kibana.crt
```
jordansissel: Possibly outside the scope of this PR, but we could remove the dependency on vault here by generating self-signed certs with `openssl` -- thoughts?

Crazybus:
> LGTM given I have only read, not tested, the security.yml. The Makefile refers to parts of Vault I don't have (and probably don't need) access to.

Thats what we have automated testing for :P You can see the results here: https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+integration-elasticsearch/12/

kibana/examples/security/Makefile
Crazybus: This is actually true for all of the testing secrets except for the license. Although for users without a license hitting the trial API would also be acceptable for testing. I'm explicitly using a real license though because I know that some rules are more relaxed when using a trial (such as no node to node encryption being required). 

I'm going to make an issue to make sure this gets done. End results is to make sure that all users and contributors can deploy the examples with minimal effort.

PR #38, opened 2019-01-14
adding securityContext/PSP support for Kibana

mszumilak:
When Pod Security Policy is in use it might be required to specify serviceAccount as well as runAsUser and fsGroups for starting the pod. Unless requested parameters are not specified pod cannot start.


kibana/values.yaml
```diff
@@ -31,6 +31,13 @@ resources:
     cpu: "1000m"
     memory: "1Gi"
 
+# If Pod Security Policy in use it may be required to specify security context as well as service account
+securityContext:
```
Crazybus: Can you rename this to podSecurityContext. Otherwise it will be a bit confusing if someone has a need to add container level security contexts too. https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

kibana/templates/deployment.yaml
```diff
@@ -19,6 +19,18 @@ spec:
         app: kibana
         release: {{ .Release.Name | quote }}
     spec:
+      {{- if .Values.securityContext }}
+      securityContext: 
+        {{- if .Values.securityContext.runAsUser }}
```
Crazybus: Instead of hardcoding two of the available options can you instead just do a `toYaml` so it is possible to set all of the [securityContext options](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#podsecuritycontext-v1-core)

kibana/values.yaml
```diff
@@ -31,6 +31,13 @@ resources:
...
+  #runAsUser: "place the user id here"
+  #fsGroup: "place the group id here"
+ 
+serviceAccount: "default"
```
Crazybus: With `{{- if .Values.serviceAccount }}` being set in the template this will always be enabled by default. Can you instead make the default that it is undefined? I would rather have users opt in for using a service account rather than accidentally giving Kibana more access than it needs.

kibana/values.yaml
mszumilak: agree, done

kibana/templates/deployment.yaml
mszumilak: done

mszumilak:
> Out of curiosity could you tell me why you need to be able to set these options? I'm mostly just curious since it is not something I have needed to do when running Kibana myself. And maybe this is something that could better be done by default if there is a more secure way to run Kibana in Kubernetes.

I have several users in my cluster with special privileges. I want to enforce using the specific, unprivileged user for running services that does not require special permissions.

> 
> Apart from the requested changes I left inline there are few other missing pieces before this is mergeable.
> 
>     1. Update the [readme](https://github.com/elastic/helm-charts/blob/master/kibana/README.md) with the two new options you have added
> 

Done

>     2. Add a [basic templating test](https://github.com/elastic/helm-charts/blob/master/kibana/tests/kibana_test.py) to make sure that the templating rules work as expected and don't get broken by future changes.
> 

Some help might be required here as I don't fully understand the script.

>     3. Seems like you still need to sign the [CLA agreement](https://www.elastic.co/contributor-agreement) before I can accept any of these changes.
> 

Done

> 
> I can help you out if needed for number 2. I realise that we don't yet have a developer guide (you are the first contributor yay!). So I'll work on making sure that these steps are included as part of the readme and the pull request template.

:+1:

Crazybus:
Thank you so much for adding this in and for explaining what the use case is!

I'll add in the template testing in a separate PR and link it from here to show you how it looks.

Crazybus:
@mszumilak Here are the tests that I added https://github.com/elastic/helm-charts/pull/44

PR #39, opened 2019-01-15
Add support for setting configuration via configmaps

Crazybus:


Crazybus:
> LGTM. Mentioning that tentative configuration changes will trigger pod updates due to the annotation checksum somewhere in the docs might be a good idea as well to avoid surprising users.

I'll add it in. I'd expect users to be surprised if changing the configuration **didn't** cause an update though.

PR #40, opened 2019-01-18
Add resources to InitContainers

rendhalver:
Current init containers don't have resource limits specified so it you are deploying this into a cluster that has a ResourceQuota specified the deploy will fail.

### DIFF
--- a/elasticsearch/Chart.yaml
+++ b/elasticsearch/Chart.yaml
@@ -4,7 +4,7 @@ maintainers:
 - email: helm-charts@elastic.co
   name: Elastic
 name: elasticsearch
-version: 6.5.4-alpha3
+version: 6.5.4-alpha4
 appVersion: 6.5.4
 sources:
   - https://github.com/elastic/elasticsearch
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -94,6 +94,8 @@ spec:
           privileged: true
         image: "{{ .Values.image }}:{{ .Values.imageTag }}"
         command: ["sysctl", "-w", "vm.max_map_count={{ .Values.sysctlVmMaxMapCount}}"]
+        resources:
+{{ toYaml .Values.initResources | indent 10 }}
       containers:
       - name: "{{ template "name" . }}"
         image: "{{ .Values.image }}:{{ .Values.imageTag }}"
@@ -101,7 +103,7 @@ spec:
         readinessProbe:
 {{ toYaml .Values.readinessProbe | indent 10 }}
           exec:
-            command: 
+            command:
               - sh
               - -c
               - |
@@ -119,7 +121,7 @@ spec:
                     fi
                     curl -XGET -s -k --fail ${BASIC_AUTH} {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}${path}
                 }
-  
+
                 if [ -f "${START_FILE}" ]; then
                     echo 'Elasticsearch is already running, lets check the node is healthy'
                     http "/"
--- a/elasticsearch/values.yaml
+++ b/elasticsearch/values.yaml
@@ -37,7 +37,7 @@ extraEnvs:
 # A list of secrets and their paths to mount inside the pod
 # This is useful for mounting certificates for security and for mounting
 # the X-Pack license
-secretMounts: 
+secretMounts:
 #  - name: elastic-certificates
 #    secretName: elastic-certificates
 #    path: /usr/share/elasticsearch/config/certs
@@ -56,6 +56,14 @@ resources:
     cpu: "1000m"
     memory: "2Gi"
 
+initResources: {}
+  # limits:
+  #   cpu: "25m"
+  #   # memory: "128Mi"
+  # requests:
+  #   cpu: "25m"
+  #   memory: "128Mi"
+
 networkHost: "0.0.0.0"
 
 volumeClaimTemplate:
@@ -67,7 +75,7 @@ volumeClaimTemplate:
 
 # By default this will make sure two pods don't end up on the same node
 # Changing this to a region would allow you to spread pods across regions
-antiAffinityTopologyKey: "kubernetes.io/hostname" 
+antiAffinityTopologyKey: "kubernetes.io/hostname"
 
 # Hard means that by default pods will only be scheduled if there are enough nodes for them
 # and that they will never end up on the same node. Setting this to soft will do this "best effort"


elasticsearch/Chart.yaml
```diff
@@ -4,7 +4,7 @@ maintainers:
 - email: helm-charts@elastic.co
   name: Elastic
 name: elasticsearch
-version: 6.5.4-alpha3
+version: 6.5.4-alpha4
```
Crazybus: Can you remove the version bump from here? Just prevents future conflicts. It would also mean that now the Kibana and Elasticsearch versions would differ once this is merged in.

elasticsearch/templates/statefulset.yaml
```diff
@@ -94,14 +94,16 @@ spec:
           privileged: true
         image: "{{ .Values.image }}:{{ .Values.imageTag }}"
         command: ["sysctl", "-w", "vm.max_map_count={{ .Values.sysctlVmMaxMapCount}}"]
+        resources:
```
Crazybus: Can you add a [small test](https://github.com/elastic/helm-charts/blob/master/elasticsearch/tests/elasticsearch_test.py) to prove that this works and that the indentation is correct. The development environment is documented in the [testing](https://github.com/elastic/helm-charts/tree/master/elasticsearch#testing) section of the readme.

elasticsearch/values.yaml
```diff
@@ -56,6 +56,14 @@ resources:
     cpu: "1000m"
     memory: "2Gi"
 
+initResources: {}
```
Crazybus: Can you also add this new variable into the readme please.

elasticsearch/Chart.yaml
desaintmartin: I see a potential problem here with the chart version / elasticsearch version tight coupling. If the version of the chart has to be the same as the elasticsearch version, it would hide important compatibility information provided by semver. This is why, in helm/charts.git, every chart have a version and an appVersion.

Example : Releasing a first version 6.5.4, then a 6.5.5 a few weeks later. One could think that the changes are only a few patches, but what if we have to break compatibility for any reason? Or add new feature?

Would you consider separating the elasticsearch version from the chart version?

Reference: https://github.com/helm/helm/blob/master/docs/charts.md#the-appversion-field

rendhalver: imageTag is a separate value and AppVerison is already set so chart version already decoupled from the image.
So I see no reason to keep chart version the same.

Crazybus: > This is the chart version.
> Chart versions should be incremented on every change.

How does this work when there are multiple pull requests open? If everyone opens a pull request and bumps the version +1 would we need to manually bump the version again and release each of the merged pull requests one by one? 

The current release plan is based on how we release all of our software (including configuration management). I'm open to the discussion of moving away from this for the helm-charts if there is a compelling reason. If you wanted to discuss further then we should continue this discussion in a dedicated issue instead. 


> It makes debugging easier when you know exactly which version of your application you are deploying.

After each Elastic release we will be bumping the version and uploading the new releases to https://helm.elastic.co. If anyone wants to test out an unreleased version they can always just install it directly from git between releases. The intention is not to re-release the chart with the current version once this changed is merged. Outside of the normal release cycle there may be hotfix releases as required but otherwise the versions will be static and match the release tags in GitHub. 

> I see a potential problem here with the chart version / elasticsearch version tight coupling. If the version of the chart has to be the same as the elasticsearch version, it would hide important compatibility information provided by semver. This is why, in helm/charts.git, every chart have a version and an appVersion.
> Example : Releasing a first version 6.5.4, then a 6.5.5 a few weeks later. One could think that the changes are only a few patches, but what if we have to break compatibility for any reason? Or add new feature?

This is part of the reason that we moved to having a unified version for all our software. The chart version is the same as the application version because we want to follow the same backwards compatibility as Elasticsearch and the rest of our software. As a user I would expect that if I install an Elasticsearch cluster using the official helm chart that I would be able to continue to use that cluster through all version upgrades. The goal with all of our configuration management tooling is to support the same [upgrade paths](https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-upgrade.html) as our software so keeping the chart and application version in sync is the only way to guarantee this. 

In other words the plan is not to make any backwards compatibility changes to the helm-charts after they have left the alpha phase. Any backwards compatibility changes will need to come with the next major release of the software that is being installed while still having a documented upgrade path. 

An example of the scenario that I'm describing if we decouple them :

* Today you install a cluster with the latest release of this helm chart
* A backwards incompatible change is made to the helm-chart that prevents you from upgrading the chart version without rebuilding your cluster
* Elasticsearch 6.6 comes out which requires (compatible) changes to the helm-chart to support running it
* A severe security vulnerability is found in your Elasticsearch version and you need to update to 6.6 but can't because there is no upgrade path. 

> imageTag is a separate value and AppVerison is already set so chart version already decoupled from the image.
> So I see no reason to keep chart version the same.

There are two reasons why this is customizable. 

1. Is to be able to run all of latest supported releases of Elasticsearch. Currently 6.5.4 and 5.6.14 which are the only versions be actively tested with the current release. Right now it is easy enough to support both major versions with the same chart but may be split sometime in the future if this becomes difficult. 
2. To allow users to specify their own customized image tags if they are using their own customized images (for example to add plugins into the docker image).

In general I would expect it to be safe to mix and match versions within minor releases but it would be untested. 

---- 

As I said at the start, I'm open to discussing making changes around how this currently works if there is a compelling reason. Managing versions this way may seem unusual if you are not familiar with the background and reasoning for how we got here.

rendhalver: Short answer to the chart version is chart release should be immutable.
`Chart releases must be immutable. Any change to a chart warrants a chart version bump even if it is only changes to the documentation.`
https://github.com/helm/charts/blob/master/REVIEW_GUIDELINES.md#immutability

A Chart is basically a package for your application.
Chart repositories are the same as package repository.
If you have a copy of that package locally helm wont download a new release because it hasn't changed.
Not incrementing the version would make it hard for helm to find a new release.
Does that explain it?

Crazybus: I think the confusion is coming from the fact that helm/charts is automatically deploying new releases whenever a version is being bumped. However that is not how our release process works. All of the charts are being released with the same version at the same time so that the versions are kept in sync with the rest of our software. 

The process looks like this:

* Pull requests are merged into master
* When it is time to release the chart versions are [bumped](https://github.com/elastic/helm-charts/pull/36/files)
* A new [release](https://github.com/elastic/helm-charts/releases) is created
* The [release](https://github.com/elastic/helm-charts/blob/master/helpers/release.py) script will create the packages and upload them to the repo. 

Even though the release process itself makes sure the package is immutable. The release script also prevents accidents from accidentally overriding an existing version.

* The release script copies the new package to the repo with `gsutil cp -n` which enforces "no-clobber" so that it cannot override an existing package. 
* The new helm repo index is generated using the `--merge` flag which makes it impossible to override or modify an existing release versions.

rendhalver: Oh I see what's going on!
Sorry for the confusion!
I will take this change out.

rendhalver: And thanks for the explanation!

elasticsearch/values.yaml
rendhalver: Done.

elasticsearch/templates/statefulset.yaml
rendhalver: Ok I will see if I can work out how that works.

rendhalver: My latest commit should address that.
I ran the test suite locally and got a pass so I think I did it right.
Let me know if I need to add more tests.

Crazybus:
jenkins test this please

elasticsearch/Chart.yaml
Crazybus: Thank you for questioning it too! Always good to have a sharp set of eyes on everything 👁 . This PR is another example that I really need to start working on a contributing guide to make this process as smooth as possible.

elasticsearch/templates/statefulset.yaml
Crazybus: Looks perfect! And a single test is more than enough for this since it is just passing through the yaml. Just want something to make sure the indentation is good and will remain that way if it ever gets refactored in the future 👍

PR #41, opened 2019-01-21
fix cluster outage, add masterService template

kimxogus:
This PR includes

1. fix cluster outage https://github.com/helm/charts/pull/10687 (https://github.com/helm/charts/issues/8785)

2. add `masterService` template in `_helpers.tpl`


elasticsearch/templates/statefulset.yaml
```diff
@@ -94,14 +96,47 @@ spec:
           privileged: true
         image: "{{ .Values.image }}:{{ .Values.imageTag }}"
         command: ["sysctl", "-w", "vm.max_map_count={{ .Values.sysctlVmMaxMapCount}}"]
+      - name: init-config
+        securityContext:
+          runAsUser: 0
+          privileged: true
+        image: "{{ .Values.image }}:{{ .Values.imageTag }}"
+        command:
+        - /bin/bash
```
desaintmartin: Maybe add some comment here to state the purpose of the initContainer?

elasticsearch/templates/statefulset.yaml
```diff
@@ -94,14 +96,47 @@ spec:
...
+
+          HOSTNAME="$(hostname)"
+          INDEX="${HOSTNAME##*-}"
+
+          ENV_VAR_PREFIX=`echo {{ template "masterService" . }}-|awk '{print toupper($0)}'|sed 's/-/_/g'`
+          HOSTVAR="${ENV_VAR_PREFIX}ANNOUNCE_${INDEX}_SERVICE_HOST"
+          HOST="${!HOSTVAR}"
+
+          if [ ! -f /usr/share/elasticsearch/config/elasticsearch.yml ]; then
+              echo "" > /usr/share/elasticsearch/config/elasticsearch.yml
```
desaintmartin: Very nitpicky, but "touch" would be more elegant.

elasticsearch/templates/statefulset.yaml
```diff
@@ -78,6 +78,8 @@ spec:
           secret:
             secretName: {{ .name }}
         {{- end }}
+        - name: config
+          emptyDir: {}
```
desaintmartin: Why should we store the configuration here instead of regenerating it at each start?

Crazybus:
@kimxogus

The issue you are linking to is referring to a different helm chart. The readinessProbes and service setup of this chart are quite different and have been designed and tested to not cause downtime during rolling upgrades and restarts. 

Were you able to reproduce the same problem with this helm chart? There is a [rolling upgrade](https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/default/rolling_upgrade.sh) script that I used when initially developing the chart to make sure it remained available. It isn't currently running as part of the automated testing (and is currently broken by the path changes to the proxy api) but this seems like a good time for it to be re-enabled to ensure this behaviour remains. 

I have just done some testing locally and wasn't able to cause any failed search requests during a rolling upgrade, killing the master pod, or doing a `kill 1` in the master pod. 

Here is how I'm testing it

```
# Deploy the default example
cd elasticsearch/examples/default
make

# Start kubectl proxy
kubectl proxy

# Run constant search requests against the service
watch -n 0.5 'curl -s --fail http://localhost:8001/api/v1/namespaces/default/services/elasticsearch-master:9200/proxy/_search'
```

If you are able to cause a failure could you please open an issue first with details on how you reproduced the problem?

elasticsearch/templates/master-announce-svc.yaml
```diff
@@ -0,0 +1,29 @@
+{{ if eq .Values.roles.master "true" }}
+{{- range $i := until (int .Values.replicas) }}
```
rendhalver: Do we really need this?
The headless service is used for service discovery and includes all members in the cluster even the unready ones
https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/service.yaml#L31

kimxogus:
@Crazybus I wrote this PR because I had cluster outage during rolling upgrade when I dived into https://github.com/elastic/elasticsearch/issues/36822, and I thought this is not fixed yet because I couldn't find changes related to connection timeout problem(https://discuss.elastic.co/t/timed-out-waiting-for-all-nodes-to-process-published-state-and-cluster-unavailability/138590/2 and https://github.com/elastic/elasticsearch/issues/36822#issuecomment-449354129). but it seems already fixed here without announce service.

This PR doesn't seem necessary any more. Thanks.

desaintmartin:
I am curious, where has it been fixed? I don't see any related change into the chart itself, is it from a new version of ES?

kimxogus:
I'm curious too. I also tested with es 6.5.3 which I used when I had cluster outage, but `curl 127.0.0.1:9200` responses immediately too which means cluster outage no longer exists.
I upgraded k8s from 1.10 to 1.11 recently, so I think some changes in k8s affected this issue.

Crazybus:
> I am curious, where has it been fixed? I don't see any related change into the chart itself, is it from a new version of ES?

> I'm curious too. I also tested with es 6.5.3 which I used when I had cluster outage

Did you reproduce the outage using this chart (elastic/helm-charts)? Or was it with the [helm/charts/elasticsearch](https://github.com/helm/charts/tree/master/stable/elasticsearch) version which you linked to in the original comment? This chart is not based on the helm/charts version so it wouldn't make sense for it to share the same issues. Nothing has been changed from `6.5.3-alpha1` to the current version that could have affected anything related to this.

desaintmartin:
What 1.11 version do you have?

kimxogus:
@Crazybus I reproduced the outage using both charts (https://github.com/elastic/elasticsearch/issues/36822#issuecomment-448818868)

@desaintmartin `1.11.7` and canal with calico `3.2.3` and flannel `0.9.0` as cni.

PR #45, opened 2019-01-24
Add configmap support for Kibana

Crazybus:
While adding in the configmap support I needed to remove the SSL
handling that was happening in the readinessProbe. 

This switches the handling of the readinessProbe to use the same
configuration as the Elasticsearch chart. With the addition of configmap
support this is now needed as it becomes very difficult to
programmatically determine whether or not https should be used and how
to locate the correct certificates and settings to connect properly.

The readinessProbe would need to properly parse the environment variables,
and the yaml file including the various different ways that the nested
yaml values can be set (like `server.ssl.enabled: true` vs `server:
<newline>ssl.enabled: true`).

So I feel that any added benefits of properly connecting via https to
verify that localhost in the docker container is actually the right
service is healthy wouldn't outweigh the added complexity this would
add. And in general I hope this will avoid any weird issues for users
who might have different SSL/certificate setups than what we are
currently testing with.


Crazybus:
jenkins test this please

Crazybus:
jenkins test this please

PR #46, opened 2019-01-25
Use a dedicated testing namespace so it can be deleted

Crazybus:
This will make sure that all resources such as pvcs are properly deleted
when destroying the testing cluster after each run.


bogeit:
It is recommended to add it in the corresponding location in the yaml file of the directory:
templates  namespace: {{ .Release.Namespace }}

now can use command:
helm install --namespace test

PR #50, opened 2019-01-29
[elasticsearch] Add configurable pod annotations

rwittrick:
Pod Annotations are required, for example, to run repository-s3 plugin.
Adding annotations (for example those available from kube2iam) allows configuration of pod level
roles in cloud providers, where the rbac is provided at the cloud provider level

Using annotations to pods allows  the configuration to be applied as in
https://www.elastic.co/guide/en/elasticsearch/plugins/current/repository-s3-repository.html#repository-s3-permissions


Crazybus:
jenkins test this please

Issue #53, opened 2019-02-07
RFC: change extraEnvs to be a map instead of a list.


jordansissel:
When using nested charts (parent/child/etc), it's often useful to add additional entries for values to pass into in child charts. However, `extraEnvs` is a list, and [helm cannot merge lists](https://github.com/helm/helm/issues/3767), but it can merge maps! (Citation: https://github.com/jordansissel/experiments/tree/master/helm/subchart-values-from-parent)

By way of example, I might have one child chart which configures Elasticsearch with my favorite default features (enable security, enable tls, etc). And I could create a new chart where I want this, plus creating a SAML realm. Because `extraEnvs` is a list, I can't just inject these SAML-specific settings. Instead, I have to copy+paste the child chart's `extraEnvs` (enable security, enable tls, etc) and append my SAML ones.

If `extraEnvs` were a map, I believe this would be solved:

Example appearance in the child described in above:
```
extraEnvs:
  xpack.security.enabled: { value: true }
  ...
  ELASTIC_PASSWORD: 
    valueFrom: { secretKeyRef: { name: elastic-credentials, key: password } }
```

And the parent chart (to add SAML, for example)
```
extraEnvs:
  xpack.security.authc.realms.saml1.type: { value: true }
  ...
```

Thoughts? I know this would deviate from Kubernetes "environment variables are defined as an list", so I'm open to discussion.

## Alternative approach

Stepping back, one alternative could be to have a general "settings" which could be used to generate a configmap instead of using environment variables for settings.

jordansissel:
I'll reopen with the  alternate approach proposal as my since that was easy to get working.

PR #54, opened 2019-02-08
Example: Inherited settings, child charts, and configuration jobs.

jordansissel:
Example usage:

```
% cd elasticsearch/examples/inheritance
% make prepare # this works around a weird bug in helm#2887

...
% make secrets
...
% kubectl create secret generic elastic-license --from-literal=license="$(cat path/to/license.json")
...
% make deploy
...
```

:clock1: :clock10: :clock4: 

```
% kubectl get pods
NAME                            READY   STATUS      RESTARTS   AGE
secure-cluster-settings-4pwdj   0/1     Completed   0          14m
secure-data-0                   1/1     Running     0          14m
secure-data-1                   1/1     Running     0          14m
secure-data-2                   1/1     Running     0          13m
secure-deploy-license-sl9g6     0/1     Completed   0          14m
secure-master-0                 1/1     Running     0          14m
secure-master-1                 1/1     Running     0          14m
secure-master-2                 1/1     Running     0          14m
```

And for the big flourish:

```
% kubectl port-forward secure-master-0 9200
...

# curl w/ the generated credentials + ca cert
% curl -u elastic:$(kubectl get secret -o json elastic-credentials | jq -r '.data["password"]' | base64 --decode) --cacert <(kubectl get secret -o json elastic-certificate-pem | jq -r '.data["elastic-certificate.pem"]' | base64 -d) https://localhost:9200/
{
  "name" : "secure-master-0",
  "cluster_name" : "secure",
  "cluster_uuid" : "eHyk_-JlQNi_Voxs8hF73w",
  "version" : {
    "number" : "6.6.0",
    "build_flavor" : "default",
    "build_type" : "tar",
    "build_hash" : "a9861f4",
    "build_date" : "2019-01-24T11:27:09.439740Z",
    "build_snapshot" : false,
    "lucene_version" : "7.6.0",
    "minimum_wire_compatibility_version" : "5.6.0",
    "minimum_index_compatibility_version" : "5.0.0"
  },
  "tagline" : "You Know, for Search"
}
```

### DIFF
--- a/elasticsearch/examples/secure-elasticsearch/templates/jobs.yaml
+++ b/elasticsearch/examples/secure-elasticsearch/templates/jobs.yaml
@@ -1,0 +1,130 @@
+{{- /* Try to only deploy maintenance/setup jobs when the role is master.
+     * This keeps us from trying to deploy the same job twice if we have
+     * separate master and data nodes defined */ -}}
+{{- if (eq .Values.elasticsearch.roles.master "true") }}
+---
+# This job deploys the Elastic license payload to the cluster once it has formed.
+# Job: {{ .Values.elasticsearch.roles.master }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-deploy-license
+spec:
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: deploy-license 
+        image: centos:7
+        command: 
+        - sh
+        - -c
+        - |
+          # Apologies for parsing JSON with shell ;)
+          # This code finds the string after the `"uid":` name.
+          # { ... "uid": "the value we want" ... }
+          license_id() {
+            tr -d '\n' | grep -Eo '"uid"\s*:\s*"[^"]+"' | awk -F'"' '{print $4}'
+          }
+
+          uid="$(echo "$license" | license_id)"
+
+          # Load curl's args into "$@"
+          set -- --max-time 10 --fail -D- --cacert /certs/elastic-certificate.pem -u "$username:$password"
+          baseurl="https://{{ .Values.elasticsearch.masterService }}:9200"
+          while true; do
+            current="$(curl "$@" -XGET "$baseurl/_xpack/license" | license_id)"
+            if [ "$?" -ne 0 ] ; then
+              echo "Failure trying to query Elasticsearch's current license."
+              sleep 10
+              continue
+            fi
+            if [ "$current" != "$uid" ] ; then
+              echo "Activating Elastic License (uid = $uid)."
+              echo "$license" | \
+
+              curl "$@" -XPUT --data-binary @- -H "Content-Type: application/json" "$baseurl/_xpack/license"
+              if [ "$?" -ne 0 ]; then
+                echo "Installing license failed. Will retry."
+                sleep 10
+              else
+                echo
+                echo "👍 License install successful."
+                break
+              fi
+            else
+              echo
+              echo "👍 License already active on this cluster."
+              break
+            fi
+          done
+        env:
+        - name: 'username'
+          valueFrom: { secretKeyRef: { name: elastic-credentials, key: username } }
+        - name: 'password'
+          valueFrom: { secretKeyRef: { name: elastic-credentials, key: password } }
+        - name: 'license'
+          valueFrom: { secretKeyRef: { name: elastic-license, key: license } }
+        volumeMounts:
+        - name: elastic-certificate-pem
+          mountPath: /certs
+      volumes:
+      - name: elastic-certificate-pem
+        secret:
+          secretName: elastic-certificate-pem
+---
+# This job pushes cluster and index settings
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-cluster-settings
+spec:
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: cluster-settings
+        image: centos:7
+        command:
+        - sh
+        - -c
+        - |
+
+          # Load curl's args into "$@"
+          set -- --max-time 10 --fail -D- --cacert /certs/elastic-certificate.pem -u "$username:$password"
+
+          set -- "$@" -XPUT -H "Content-Type: application/json"
+          baseurl="https://{{ .Values.elasticsearch.masterService }}:9200"
+
+          try() {
+            while ! "$@" ; do sleep 10; done
+          }
+
+          {{- if .Values.elasticsearch.settings.cluster }}
+          # Deploy cluster settings
+          settings='{ "persistent": {{ .Values.elasticsearch.settings.cluster | toJson }} }'
+          echo "Cluster settings: " "$settings"
+          try curl "$@" "$baseurl/_cluster/settings" --data-binary "$settings"
+          {{- end }}
+
+          {{- if .Values.elasticsearch.settings.index }}
+          # Deploy _all indices settings
+          settings='{ "settings": {{ .Values.elasticsearch.settings.index | toJson }} }'
+          echo "Index (_all) settings:" "$settings"
+          try curl "$@" "$baseurl/_all/_settings" --data-binary "$settings"
+          {{- end }}
+        env:
+        - name: 'username'
+          valueFrom: { secretKeyRef: { name: elastic-credentials, key: username } }
+        - name: 'password'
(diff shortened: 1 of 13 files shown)


jordansissel:
One note: This is somewhat copied from an existing chart I'm using already, so the Makefile has some extra bits lying around (kibana and nginx secret generation). If you like the direction of this PR, I'm happy to remove those to clean it up a bit.

elasticsearch/examples/secure-elasticsearch/templates/jobs.yaml
```diff
@@ -0,0 +1,130 @@
+{{- /* Try to only deploy maintenance/setup jobs when the role is master.
+     * This keeps us from trying to deploy the same job twice if we have
+     * separate master and data nodes defined */ -}}
+{{- if (eq .Values.elasticsearch.roles.master "true") }}
+---
+# This job deploys the Elastic license payload to the cluster once it has formed.
+# Job: {{ .Values.elasticsearch.roles.master }}
+apiVersion: batch/v1
+kind: Job
```
Crazybus: I'm trying to imagine how this is supposed to look like. The `command` field in a job spec is immutable so it can't be updated or changed. If you did want to have some kind of "run the job when it changes or gets updated" it would need to do a trick like adding the checksum of the file into the metadata.name to force a new resource to be created.  I think helm would also then cleanup the previous job but I'm not too sure.

elasticsearch/values.yaml
```diff
@@ -18,6 +18,24 @@ minimumMasterNodes: 2
...
+settings:
+  # These settings become part of elasticsearch.yml on the node.
+  # It should be structured, not a text block.
+  node: {}
+    # Example, to enable Elasticsearch Security
+    #xpack.security.enabled: true
+
+  # These settings are published using the /_cluster/settings endpoint
+  # TODO(sissel): Not implemented yet.
+  cluster: {}
```
Crazybus: Is there a specific reason to split these out into cluster and index? Rather than just having a generic way to set settings that need to be done via the API? I ask this because the two settings I'm currently setting having the paths `/_snapshot/default` and `_all/_settings` which would not be covered by these. Instead just having a generic way to specify the path and the payload you want feels like it would work nicely. 

```
settings:
- path: /_all/_settings
  data:
    settings:
      index.unassigned.node_left.delayed_timeout: 5m
```

Which would just generate a curl command with the json payload. (pseudo code)

```
curl -XPUT 'http://elasticsearch:9200/{{ path }}' -H "Content-Type: application/json" -d "{{ toJson data }}"
```

Crazybus:
> % kubectl create secret generic elastic-license --from-literal=license="$(cat path/to/license.json")

Just an FYI that this can be done with `kubectl create secret generic elastic-license --from-file=license.json`

elasticsearch/examples/secure-elasticsearch/templates/jobs.yaml
jordansissel: In practice, helm does poorly with Jobs because it attempts to push a new job config (if the job changes) and is rejected by kubernetes. The only solution is to delete the job, then `helm install` (or upgrade) again. It's annoying.

I think a Job is wrong for this, but I can't think of any other more appropriate solution.

elasticsearch/templates/statefulset.yaml
```diff
@@ -189,6 +187,11 @@ spec:
             subPath: {{ .subPath }}
             {{- end }}
           {{- end }}
+          {{- if .Values.settings.node }}
+          - name: esconfig
```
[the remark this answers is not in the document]
jordansissel: This is true for the current setting also, is it not? If you have:

```
esConfig:
  elasticsearch.yml: ...
  log4j2.properties: ...
```

You'll get two `esconfig` definitions from this code below:

```
   {{- range $path, $config := .Values.esConfig }}
          - name: esconfig
```

elasticsearch/values.yaml
```diff
@@ -18,6 +18,24 @@ minimumMasterNodes: 2
 
 esMajorVersion: 6
 
+# Settings for your nodes, clusters, and indices.
+settings:
+  # These settings become part of elasticsearch.yml on the node.
+  # It should be structured, not a text block.
```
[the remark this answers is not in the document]
jordansissel: > to be able to merge these settings

My use case is less about node types and more about different clusters. The scenario I'm envisioning is this:

* There is a baseline configuration (security enabled, https, etc)
* Individual clusters may have slight deviations

For work, we have a single repo housing the configuration for multiple independent clusters, and I want to reduce copy/pasting between clusters. For example, we have a main logs cluster that has a staging implementation which is nearly identical except for sizing (fewer data nodes, smaller disk, etc). What brought me to this merging solution is that I want to explore using SAML, and for that, I want to inherit my base configuration (security enabled, https, change default node_left timeout to 5m, etc) and append SAML settings to it. This cannot be done with extraEnvs (see #53) because it is a list, so either way the configuration (via esConfig or via extraEnvs) would be a full  copy+paste resulting in duplication as well as what I see as some maintenance hazards (diffing similar-purpose dev/staging/prod clusters becomes somewhat annoying, extending the baseline configuration is copy+paste instead of just adding specific additional settings, etc).

Does this make sense?

elasticsearch/templates/statefulset.yaml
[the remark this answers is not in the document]
jordansissel: should the `name` in the current code be based on the path? like `- name: esconfig-{{ $path }}` (assuming we use appropriate string functions to turn `$path` into a valid name.

elasticsearch/examples/secure-elasticsearch/templates/jobs.yaml
jordansissel: Thinking about it more, this is probably best done by a controller or operator-style service which watches the configmap(s) and ships new settings to Elasticsearch when necessary. Such a thing may be out of scope for this helm-chart, though I would use it in production. Alternately, instead of a job, maybe a pod definition that is basically `while true; do sync configmap to elasticsearch if it changes; done` ?

elasticsearch/values.yaml
jordansissel: I am +1 on what you suggest, though I think this would mean we should separate API-driven-settings from `elasticsearch.yml`. It would feel weird to mix them:

```
settings:
  node: 
    # elasticsearch.yml contents here
  /_all/settings:
    index.unassigned.node_left.delayed_timeout: 5m
  ...
```

For your example, you used a list-of-maps, but lists cannot be merged in Helm across charts, and having two `/_all/_settings` entries doesn't make sense to me. Could it be a map of `path: settings` as I have in this comment?

elasticsearch/values.yaml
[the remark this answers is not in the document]
jordansissel: To put a concrete example:

In all my clusters, I'm likely to want the same baseline config (as said, https, security enabled, etc). In the near future, I'm going to explore SAML, and based on my reading so far, each cluster will need it's own SAML ACS per cluster, which means each cluster will need several settings unique per cluster: Under xpack.security.authc.realms.[name], you'd have `sp.entity_id` and `sp.acs` -- different for each cluster. It's possible these could be `cluster` settings shipped with `/_cluster/settings` api, but I haven't tried that.

Roughly, I'm usually looking for ways to reduce maintenance errors and such, and I feel like extending the configuration (vs copy-paste-modify'ing it) is one way to achieve that.

jordansissel:
In doing some further experimentation, I find that today's chart lets me extend at least once by mixing env vars and elasticsearch.yaml:

My `child` chart (describing the baseline configuration for all clusters) uses env vars to enable things like security. Then the `parent` chart (the specific instance of a cluster) can extend by defining `esConfig`'s `elasticsearch.yml`

In a way, it's not ideal because the parent and child are required to define settings with two different syntaxes, but it does work at least.

elasticsearch/values.yaml
jordansissel: Thinking more about this, yeah I definitely agree on paths, since these could be used to deploy a whole bunch of things I've had to deal with lately: watches, security roles, cluster settings, _all index settings, etc.

jordansissel:
> Just an FYI that this can be done with kubectl create secret generic elastic-license --from-file=license.json

The files downlaoded from the license portal are unpredictable things like `foo-bar-baz-UUID.json` and I think `--from-file` always uses the filename as the secret's key? This is mainly why I used `--from-literal=...$(cat ...)`

Issue #60, opened 2019-02-12
[elasticsearch] Allow for configurable health status value in readinessProbe


rwittrick:
In the elasticsearch statefulset readinessProbe command, the cluster health status has a hardcoded value of `green`.  It's not always the case that, as with coordinator nodes, green is the minimum acceptable state for cluster health.

tgruenert:
I´ve run in the same issue at this moment. It would be nice to configure the required level within values. "green" can be default to be backwards compatible

```
readinessProbe.level: green
```

rwittrick:
Doing that would add level to the configuration of the readinessProbe itself.  A couple of suggestions for the values file are:
1. Put the value at the top level outside of the readinessProbe
```
minClusterHealthStatus: green
```
2. Move the readinessProbe configuration fields under config, and the status field under readinessProbe
```
readinessProbe:
  config:
    failureThreshold: 3
    initialDelaySeconds: 10
    periodSeconds: 10
    successThreshold: 3
    timeoutSeconds: 5
  minClusterHealthStatus: green
```
3. Similar to above, but put the status value more explicitly in the `command` key to differentiate it from k8s config
```
readinessProbe:
  config:
    failureThreshold: 3
    initialDelaySeconds: 10
    periodSeconds: 10
    successThreshold: 3
    timeoutSeconds: 5
  command:
    minClusterHealthStatus: green
```

I'll work on a PR using option 3, for now

rwittrick:
Actually option 3 above would be a breaking change for anyone that has already implemented values for the readinessProbe.  The safest option is 1.

tgruenert:
thank you for your feedback. after thinking about, my first suggestion was not the best one.
because i needed a quick fix for my problem, i also forked the project and did a change simular to your option 3. 
it would be nice to see your PR for tis topic. thank you!

rwittrick:
@tgruenert - see https://github.com/elastic/helm-charts/pull/62 :)

jordansissel:
> the cluster health status has a hardcoded value of `green`

My reading of the readinessProbe is that it checks for green *once* and once green only checks the node's root path `GET /` after that. It doesn't require the cluster to always be green.

I know it doesn't require green all the time because my cluster goes yellow (normal) during maintenance and none of my Elasticsearch nodes fail their readinessProbe.

rwittrick:
[...]

PR #62, opened 2019-02-14
issue-60: Add a configurable health status to the readinessProbe command

rwittrick:


elasticsearch/values.yaml
```diff
@@ -118,6 +118,9 @@ readinessProbe:
   successThreshold: 3
   timeoutSeconds: 5
 
+# https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html#request-params wait_for_status
+minClusterHealthStatus: green
```
Crazybus: There are so many custom paramters here that people might want to change other than `wait_for_status`. How about making this more generic and having it be something like:

```
clusterHealthCheckParams: wait_for_status=green&timeout=1s
```

This is going to make this a lot more maintainable if users want to modify the timeout...or add other settings, or if a feature release of Elasticsearch changes the parameters.

elasticsearch/templates/statefulset.yaml
```diff
@@ -109,14 +109,14 @@ spec:
         image: "{{ .Values.image }}:{{ .Values.imageTag }}"
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
         readinessProbe:
-{{ toYaml .Values.readinessProbe | indent 10 }}
+{{ toYaml .Values.readinessProbe.config | indent 10 }}
```
Crazybus: Did you mean to leave this change in? Reading through #60 makes me believe this was leftover from testing a different approach. In any case this will fail with the current default values files and tests.

elasticsearch/values.yaml
rwittrick: I've updated to that format.

Crazybus:
jenkins test this please

PR #67, opened 2019-02-26
Bump 7 alpha testing to 7 beta 1

Conky5:
`7.0.0-beta1` is out, let's try it.


helpers/matrix.yml
```diff
@@ -5,12 +5,12 @@ ES_SUITE:
   - default
   - multi
   - security
-  - 7-alpha
+  - 7-beta
```
Crazybus: I think we can just rename this to 7.x already now that it is in the beta phase.

kibana/templates/deployment.yaml
```diff
@@ -64,6 +64,8 @@ spec:
         env:
           - name: ELASTICSEARCH_URL
             value: "{{ .Values.elasticsearchURL }}"
+          - name: ELASTICSEARCH_HOSTS
+            value: "{{ .Values.elasticsearchURL }}"
```
Crazybus: As this is supported in 6.6.0 and elasticsearchURL won't be supported in 7.x I think it makes the most sense to add a new `elasticsearchHosts` variable which is commented out by default. And updating both environment variables to only be added if they have been defined. 

Once 7.x is the stable release we can switch them around. So that would be making `elasticsearchHosts` the default with `elasticsearchURL` being commented out so that the chart would still work with releases before 6.6.0. 

This helps to keep things in-sync with Kibana. If Kibana 7.x doesn't let you set elasticsearchURL then we shouldn't either and could cause some confusion.

helpers/matrix.yml
Conky5: `7.0.0` is in beta, `7.x` is a branch targeting `7.1.0` in the elastic stack projects. So I think `7.x` may cause some confusion if it is used to target `7.0.0-beta1` images.

Perhaps `7.0.x` could go here?

Conky5: Pushed rename to `7.0.x`.

kibana/templates/deployment.yaml
Conky5: That sounds reasonable to me, pushed in https://github.com/elastic/helm-charts/pull/67/commits/e2ba0a28747ea8f46397cd14781b9c106dc439c3.

kibana/README.md
```diff
@@ -24,7 +24,8 @@ This helm chart is a lightweight way to configure and run our official [Kibana d
...
 | ------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
-| `elasticsearchURL`        | The URL used to connect to Elasticsearch.                                                                                                                                                                                                          | `http://elasticsearch-master:9200`                                                                                        |
+| `elasticsearchURL`        | The URL used to connect to Elasticsearch. Supported previous to Kibana 6.6.0.                                                                                                                                                                      | `http://elasticsearch-master:9200`                                                                                        |
```
Crazybus: It's still supported but only deprecated for < 7. This makes it sound like it won't work at all in 6.6.0 which is not the case.

Conky5: 👍  Pushed https://github.com/elastic/helm-charts/pull/67/commits/2c24b8c62a8d4e9cf30e0defda3c9ad8aeaa9167

Conky5:
Snuck in a `:TableFormat` of the markdown tables in https://github.com/elastic/helm-charts/pull/67/commits/95678e0913324ba122039ff79db0f948824e7d55, hope that is alright, otherwise I can revert it.

Crazybus:
> Snuck in a :TableFormat of the markdown tables in 95678e0, hope that is alright, otherwise I can revert it.

Fine with me! I use it as well but its pretty hard to keep it up with other contributors.

Crazybus:
Thanks for fixing this up @Conky5! Plenty more work in the backlog if you want some more ;)

PR #65, opened 2019-02-26
Remove kubernetes 1.10 from ci testing

Conky5:
Version `1.10.x` is no longer available as of February 18th 2019 [1].

This is causing failures in ci [2].

[1]:
https://cloud.google.com/kubernetes-engine/docs/release-notes#february-18-2019
[2]:
https://devops-ci.elastic.co/job/elastic+helm-charts+master+cluster-creation/KUBERNETES_VERSION=1.10,label=docker&&virtual/109/console

PR #75, opened 2019-03-11
Add all currently supported GKE versions

Crazybus:
Also added some reliability improvements to reduce test flakiness

* Retry goss validate every 5 seconds for up to 30 seconds. (was occasionally failing on Kibana tests)
* Retry helm init 5 times

### DIFF
--- a/helpers/matrix.yml
+++ b/helpers/matrix.yml
@@ -13,4 +13,6 @@ KIBANA_SUITE:
   - 7.0.x
   - 5.x
 KUBERNETES_VERSION:
+  - '1.10'
   - '1.11'
+  - '1.12'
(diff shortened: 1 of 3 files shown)


helpers/matrix.yml
```diff
@@ -13,4 +13,6 @@ KIBANA_SUITE:
   - 7.0.x
   - 5.x
 KUBERNETES_VERSION:
+  - '1.10'
```
Conky5: I actually removed this recently https://github.com/elastic/helm-charts/pull/65/files as GKE no longer has it available.

Crazybus: It seems like they added it back in again or it was temporarily unavailable. As proof you can see it is passing in this pull request testing job https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+cluster-creation/93/ .

Conky5: Hmm interesting.

Crazybus: https://cloud.google.com/kubernetes-engine/docs/release-notes#version-updates

> 1.10.12-gke.7 - This version is being made available again after being previously removed.
> 1.10.12-gke.9

Pretty weird really /shrug

PR #76, opened 2019-03-12
Add 7.1.0 snapshot testing

Conky5:
Add tests for `7.1.x` for testing `7.x` branch development.

### DIFF
--- a/helpers/matrix.yml
+++ b/helpers/matrix.yml
@@ -6,11 +6,13 @@ ES_SUITE:
   - multi
   - security
   - 7.0.x
+  - 7.1.x
   - 5.x
 KIBANA_SUITE:
   - default
   - security
   - 7.0.x
+  - 7.1.x
   - 5.x
 KUBERNETES_VERSION:
   - '1.10'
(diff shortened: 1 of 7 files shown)


helpers/matrix.yml
```diff
@@ -6,11 +6,13 @@ ES_SUITE:
   - multi
   - security
   - 7.0.x
+  - 7.1.x
```
Crazybus: I'm ok with snapshot versions being included as examples but it doesn't make sense for them to be added to the list of test suites. This matrix is used for pull request and master branch testing. It's not very nice if someones pull request randomly starts failing because of a change in one of the snapshots. The same thing applies for the master job which we try to keep as green as possible and require a passing build from before releasing a new helm chart version. As a general rule only actual releases with static docker image tags should end up in this list. 

If we want to catch breaking changes in snapshot versions then creating a second periodic job only testing snapshots would be the way to go.

Conky5: I agree, I will get to this eventually.

PR #77, opened 2019-03-13
[elasticsearch] master service template

kimxogus:
This pr adds `masterService` template which automatically set master service name.


elasticsearch/Chart.yaml
```diff
@@ -4,7 +4,7 @@ maintainers:
 - email: helm-charts@elastic.co
   name: Elastic
 name: elasticsearch
-version: 6.6.0-alpha1
+version: 6.6.0-alpha2
```
Crazybus: Could you remove the version bump please? We don't automatically deploy new releases each time a pull request is merged like with helm/charts. Instead all charts are bumped and released at the same time to keep them all in sync. I went into more detail in https://github.com/elastic/helm-charts/pull/40#discussion_r249422024 if you are interested in the reasoning why.

Crazybus:
jenkins test this please

elasticsearch/Chart.yaml
kimxogus: removed

Issue #63, opened 2019-02-17
Slow re-election when elected master pod is deleted


andreykaipov:
First of all - thank you guys for the chart!

I was playing around with the multi-node example and experienced some odd behavior. Here's how I'm reproducing the issue.

After the [multi example](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples/multi) is deployed, open up the multi-data service to your local in one terminal:
```
$ kubectl port-forward service/multi-data 9200
``` 

Watch the call to `/_cat/master` in another terminal:
```
$ watch -n1 time curl -s http://localhost:9200/_cat/master?v
```

In a third terminal, whoever the elected master is, delete them:
```
$ kubectl delete pod multi-master-0
```

The API call in the second terminal will now hang. After 30 seconds, the request will timeout and we might see the following error for a split second:
```
{"error":{"root_cause":[{"type":"master_not_discovered_exception","reason":null}],"type":"master_not_discovered_exception","reason":null},"status":503}
```

Soon after, the cluster recovers and the API call from the second window starts responding again. Here are the logs off another master node before and after the re-election:
```
[2019-02-17T01:49:03,736][INFO ][o.e.d.z.ZenDiscovery     ] [multi-master-1] master_left [{multi-master-0}{KZPjmKZtSf2LGV-IyvtfOg}{Es2wrbyiQdWz5CnT4V5wkA}{10.40.1.14}{10.40.1.14:9300}{ml.machine_memory=2147483648, ml.max_open_jobs=20, xpack.installed=true, ml.enabled=true}], reason [shut_down]
[2019-02-17T01:49:03,736][WARN ][o.e.d.z.ZenDiscovery     ] [multi-master-1] master left (reason = shut_down), current nodes: nodes:
   {multi-master-0}{KZPjmKZtSf2LGV-IyvtfOg}{Es2wrbyiQdWz5CnT4V5wkA}{10.40.1.14}{10.40.1.14:9300}{ml.machine_memory=2147483648, ml.max_open_jobs=20, xpack.installed=true, ml.enabled=true}, master
   {multi-data-0}{Ndz2WGGiSz6Y1XO1tWyWRw}{nPoZagPdRr2Tq45BPAkh_g}{10.40.2.13}{10.40.2.13:9300}{ml.machine_memory=2147483648, ml.max_open_jobs=20, xpack.installed=true, ml.enabled=true}
[...]

PR #82, opened 2019-03-19
custom command for elasticsearch master nodes

kimxogus:
#63

https://github.com/elastic/helm-charts/issues/63#issuecomment-471096950

As fix from upstream will be in 7.1, this change seems to be needed now.


elasticsearch/templates/statefulset.yaml
```diff
@@ -108,6 +108,20 @@ spec:
...
+          - -c
+          - |
+            # https://github.com/elastic/helm-charts/issues/63#issuecomment-471096950
+            # Run ES as a background task, and forward SIGTERM to it, then wait for it to exit
+            trap 'kill $(jobs -p)' SIGTERM
+            /usr/local/bin/docker-entrypoint.sh elasticsearch &
+            wait
+            # now keep the pod alive for 30s after ES dies so that we will refuse connections from
+            # the new master rather than them needing to time  out
+            sleep 30
```
Crazybus: Why 30 seconds? How do you know that this will be enough time? Magic numbers are not normally a good sign and should only be used if there is no way to determine how long we should wait. 

Luckily it is possible to detect that a new master has been elected as I commented in https://github.com/elastic/helm-charts/issues/63#issuecomment-465081133. This is what I would prefer to see implemented. As it doesn't rely on magic numbers and it also means that we can continue to use the default entrypoint for the docker images.

elasticsearch/templates/statefulset.yaml
```diff
@@ -108,6 +108,20 @@ spec:
...
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
+        {{- if eq .Values.roles.master "true" }}
+        command:
+          - /bin/sh
+          - -c
+          - |
+            # https://github.com/elastic/helm-charts/issues/63#issuecomment-471096950
+            # Run ES as a background task, and forward SIGTERM to it, then wait for it to exit
+            trap 'kill $(jobs -p)' SIGTERM
+            /usr/local/bin/docker-entrypoint.sh elasticsearch &
```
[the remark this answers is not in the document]
kimxogus: fixed argument passing to es.

elasticsearch/templates/statefulset.yaml
kimxogus: added `sleepSecondsOnTermination` value and set `30` as default value.

kimxogus:
@Crazybus Can you please review?

I've been running es cluster with this command option for about 2 weeks and working fine with environment variables and additional args.

elasticsearch/templates/statefulset.yaml
Crazybus: Making it configurable is certainly better than hard coding it. But this doesn't address my original feedback from https://github.com/elastic/helm-charts/pull/82#discussion_r266901909. Why is this 30 seconds, and why is that long enough to prevent the problem defined in #63. Is this calculated based on something?

Crazybus:
jenkins test this please

elasticsearch/templates/statefulset.yaml
kimxogus: I just copied and pasted https://github.com/elastic/helm-charts/issues/63#issuecomment-471096950 . That's where 30 seconds came from.  In my test environment, around 10 seconds were sufficient.

elasticsearch/templates/statefulset.yaml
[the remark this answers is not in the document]
kimxogus: I found that broken usage of env vars. Maybe I couldn't found this problem because of my custom `elasticsearch.yml` file.

Fixing broken env vars is beyond my knowledge of bash scripts, hope someone to fix this issue

andreykaipov:
I feel like maintaining a workaround in this chart isn't the right solution for this. A better approach might be to expose the `command` and `args` of the container as values so users can change the entrypoint themselves.

Additionally, just like we have an `extraInitContainers` value, an `extraContainers` value would be handy so users can implement the sidecar workaround instead of overriding the entrypoint. The ability to run sidecar containers is useful in any case. 😄

PR #86, opened 2019-03-25
Add migration guide for migrating from the community helm/charts repo

Crazybus:
Related: https://github.com/helm/charts/issues/10543

This is the first step in making it possible for users of the
helm/charts version to migrate to this chart. The other
part of this is writing up a guide to help users convert their existing
values files into something compatible with the new chart.

### DIFF
--- a/elasticsearch/examples/migration/README.md
+++ b/elasticsearch/examples/migration/README.md
@@ -83,4 +83,4 @@ The process will involve a re-sync and a rolling restart of all of your data nod
   ```
   kubectl scale deployment my-release-elasticsearch-client --replicas=0
   ```
-* The migration should now be complete. After verifying that everything is working correctly you can cleanup and leftover resources from your old cluster.
+* The migration should now be complete. After verifying that everything is working correctly you can cleanup leftover resources from your old cluster.
--- a/elasticsearch/examples/migration/README.md
+++ b/elasticsearch/examples/migration/README.md
@@ -47,7 +47,7 @@ The process will involve a re-sync and a rolling restart of all of your data nod
   kubectl scale statefulsets my-release-elasticsearch-data --replicas=0
   ```
 * Wait for the cluster to be green again.
-* OK. We now have all data nodes running in the new cluster. Time to replace the masters by firstly scaling down the masters from 3 to 2. Between each step make sure to wait for the cluster to become green again. During this process we will always make sure to keep at least 2 master nodes.
+* OK. We now have all data nodes running in the new cluster. Time to replace the masters by firstly scaling down the masters from 3 to 2. Between each step make sure to wait for the cluster to become green again, and check with `curl localhost:9200/_cat/nodes` that you see the correct amount of master nodes. During this process we will always make sure to keep at least 2 master nodes as to not lose quorum.
   ```
   kubectl scale statefulsets my-release-elasticsearch-master --replicas=2
   ```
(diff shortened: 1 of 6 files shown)


Crazybus:
cc @rendhalver. I'm trying to request a review but GitHub isn't letting me.

elasticsearch/examples/migration/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+This example will be using the default values for the existing helm/charts release and for the elastic helm-charts release. If you have changed any of the default values then you will need to first make sure that your values are configured in a compatible way before starting the migration. 
+
+The process will involve a re-sync and a rolling restart of all of your data nodes. Therefore it is important to disable shard allocation and perform a synced flush like you normally would during any other rolling upgrade. See the [rolling upgrades guide](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/rolling-upgrades.html) for more information.
+
+* The default image for this chart is `docker.elastic.co/elasticsearch/elasticsearch` which contains the default distribution of Elasticsearch with a [basic license](https://www.elastic.co/subscriptions). Make sure to update the `image` and `imageTag` values to the correct Docker image and Elasticsearch version that you currently have deployed.
+* Convert your current helm/charts configuration into something that is compatible with this chart.
```
tylerjl: Could you maybe include a sub-bullet point here with an illustrative example of what sort of settings this might include? I assume this is something like `cluster.config`, but a broad statement about whether this applies things like `persistence` or `serviceType` might be helpful to illustrate what sort of settings should be checked.

elasticsearch/examples/migration/README.md
```diff
@@ -0,0 +1,86 @@
...
+  ```
+* Deploy the client nodes
+  ```
+  make client
+  ```
+* Update any processes that are talking to the existing client nodes and point them to the new client nodes. Once this is done you can scale down the old client nodes
+  ```
+  kubectl scale deployment my-release-elasticsearch-client --replicas=0
+  ```
+* The migration should now be complete. After verifying that everything is working correctly you can cleanup and leftover resources from your old cluster.
```
tylerjl: `s/you can cleanup and leftover/you can cleanup leftover/`?

elasticsearch/examples/migration/README.md
Crazybus: Agreed. Although the plan is to handle this in a separate PR which I noted in the PR description:

> This is the first step in making it possible for users of the
helm/charts version to migrate to this chart. The other
part of this is writing up a guide to help users convert their existing
values files into something compatible with the new chart.

elasticsearch/examples/migration/README.md
Crazybus: Fixed. Sorry English is not my first language, I grew up speaking Strayan.

Crazybus:
> I think like this is useful/important enough to potentially call out in the top-level README

Good call! Added it in. Will save having someone open a GitHub issue asking how to do it :P 

> One aspect (that I don't think will be an issue) but I still wanted to mention crossed my mind is syncing master node state. In between scaling down old masters and scaling up new ones, the new masters will need to a) join the quorum and b) sync the cluster state in order to commit and save all non-index segment data (templates, etc.). Ensuring that the cluster is green and stable in-between scaling events probably captures this, but I'm not an expert about what this behavior is so I wanted to explicitly mention it to ensure that you're also in agreement that it is probably addressed in the scale/check cycle you mention in these steps.

Thanks for calling this out. It seems that a similar process has been suggested in [discuss](https://discuss.elastic.co/t/adding-new-master-nodes-to-old-cluster/155887/2) before and was also blessed by [zen master David](https://discuss.elastic.co/t/adding-new-master-nodes-to-old-cluster/155887/3). One subtle difference is that this guide specifically calls out to check that you have 3 master eligible nodes after each step. I'll add that in to be on the safe side.

rendhalver:
Thank you @Crazybus this is so very awesome!
Sorry I wasn't able to check this out earlier.
This is exactly the the thing we need to keep this going. :)
Awesome work!

PR #94, opened 2019-04-12
bump es version, remove default storageClassName

kimxogus:
This pr
1. Bump default es version to `6.7.1`
2. Removes `standard` storageClassName by default as this value will be default storage class if empty(you don't need to specify this). In aws environment, default storage class is generally `gp2`, so `volumeClaimTemplate` should be always overridden.


kimxogus:
should be `7.0.0`?

z0lope0z:
👍 trying to helm install on new AWS EKS cluster and getting the error

persistentvolume-controller  storageclass.storage.k8s.io "standard" not found

elasticsearch/Chart.yaml
```diff
@@ -4,8 +4,8 @@ maintainers:
 - email: helm-charts@elastic.co
   name: Elastic
 name: elasticsearch
-version: 6.6.2-alpha1
```
Crazybus: Can you remove this version bump please. All versions of the charts are bumped and released at the same time (just like with all of the elastic stack). I went into more details in this PR if you are interested in the reasons why: https://github.com/elastic/helm-charts/pull/40#discussion_r249422024

elasticsearch/values.yaml
```diff
@@ -71,7 +71,6 @@ networkHost: "0.0.0.0"
 
 volumeClaimTemplate:
   accessModes: [ "ReadWriteOnce" ]
-  storageClassName: "standard"
```
Crazybus: Awesome! This is much better. Can you also update the defaults in the readme to match https://github.com/elastic/helm-charts/tree/master/elasticsearch#configuration

Crazybus:
jenkins test this please

elasticsearch/values.yaml
kimxogus: updated readme

elasticsearch/Chart.yaml
kimxogus: reverted chart version bump

elasticsearch/Chart.yaml
```diff
@@ -5,7 +5,7 @@ maintainers:
   name: Elastic
 name: elasticsearch
 version: 6.6.2-alpha1
-appVersion: 6.6.2
+appVersion: 6.7.1
```
Crazybus: Something went wrong with the revert as the version bump is still here.

elasticsearch/README.md
```diff
@@ -55,14 +55,14 @@ If you currently have a cluster deployed with the [helm/charts stable](https://g
...
-| `imageTag`                 | The Elasticsearch docker image tag                                                                                                                                                                                                                                                                                         | `6.6.2`                                                                                                                   |
+| `imageTag`                 | The Elasticsearch docker image tag                                                                                                                                                                                                                                                                                         | `6.7.1`                                                                                                                   |
```
Crazybus: This one should be reverted too.

elasticsearch/values.yaml
```diff
@@ -43,7 +43,7 @@ secretMounts: []
 #    path: /usr/share/elasticsearch/config/certs
 
 image: "docker.elastic.co/elasticsearch/elasticsearch"
-imageTag: "6.6.2"
+imageTag: "6.7.1"
```
Crazybus: This one too

PR #95, opened 2019-04-12
Update documentation links

Conky5:
Update documentation links now that 7.0.0 has been released.


elasticsearch/README.md
```diff
@@ -39,10 +39,10 @@ This helm chart is a lightweight way to configure and run our official [Elastics
...
-| `masterService`            | Optional. The service name used for [discovery.zen.ping.unicast.hosts](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-discovery-zen.html#unicast) to connect to the masters. You only need to set this if your master `nodeGroup` is set to something other than `master`                         | ``                                                                                                                        |
+| `masterService`            | Optional. The service name used for [discovery.zen.ping.unicast.hosts] to connect to the masters. You only need to set this if your master `nodeGroup` is set to something other than `master`. Deprecated but currently supported in versions >= 7, replaced by [discovery.seed_hosts].                                   | ``                                                                                                                        |
```
Crazybus: > Deprecated but currently supported

This is a bit confusing since it sounds like the `masterService` value is deprecated. In reality it is just an implementation detail. Instead we should update the chart to use `discovery.seed_hosts` when the version is 7 or greater. Deprecated makes it sound like the user should stop using this value which is not the case here.

elasticsearch/README.md
```diff
@@ -101,7 +104,7 @@ make
 
 A cluster with X-Pack security enabled
 
-* Generate SSL certificates following the [official docs]( https://www.elastic.co/guide/en/elasticsearch/reference/6.3/configuring-tls.html#node-certificates)
+* Generate SSL certificates following the [official docs]( https://www.elastic.co/guide/en/elasticsearch/reference/6.4/configuring-tls.html#node-certificates)
```
Crazybus: Why 6.4? TBH I have no idea why I linked to 6.3 but the 6.7 link seems to be the same too. Keeping them all the same will make it easier to do `s/6.7/6.8/g` on the file.

elasticsearch/README.md
Crazybus: Instead I think it is better to just update the wording to remove any implementation details and link to `#clustering-and-node-discovery` which explains it in more details. 

So something like:

Optional. The service name used to connect to the masters. You only need to set this if your master `nodeGroup` is set to something other than `master`. See #clustering-and-node-discovery for more information.

elasticsearch/README.md
Conky5: 6.4 appeared to be the last version that referenced x-pack for the linked documentation
> elasticsearch-certutil has been included with x-Pack

I thought this section was explicitly for x-pack security, but it is not, also this does not make sense since x-pack was opened started with 6.3 anyway.

I've updated to 6.7 😄

elasticsearch/README.md
Conky5: Updated with your suggested description.

Issue #93, opened 2019-04-12
README.md needs an update


a139786:
heml install line still has 6.6.2-alpha1

a139786:
Of course, that should read helm vs heml

jethr0null:
6.6.2-alpha1 is still correct as it is the current version of the chart itself. The image value can be overridden to specify the desired image version (though other configuration values may also need to be changed depending on which version is specified). 

That said, we have plans to bump the chart version to 7.0.0-alpha1 on Monday. In this version of the chart, the default image will be the version 7.0.0 (GA) image and the README will also be updated at that point. 

Thanks so much!

Issue #91, opened 2019-04-11
Bump chart to version 7.0.0


jethr0null:
@Crazybus can you please bump the chart version to 7.0.0?

PR #96, opened 2019-04-16
Switch to 7.x as the default version

Crazybus:
Closes: #91 
Closes: #93


helpers/bumper.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python2
+import re
+import os
+import glob
+import subprocess
+import fileinput
+
+os.chdir('../')
```
Conky5: Totally not me: _blindly executes `./helpers/bumper.py`_
Totally not me: ooooh 🤦‍♂️ 

Consider some guardrails to ensure this gets executed at the right level.

Crazybus: 🤦‍♂️ Yeah I'll make sure that it also goes to the parent directory from where the file is in a separate PR. The chance of the globs actually matching something is pretty low though...but still.

Conky5: Yeah definitely user error, it only touched README's for me across elastic's repositories.

Issue #103, opened 2019-04-18
Kibana: Readiness vs basePath


johnmarcou:
Hello,

When customising the basePath, Kibana expect according request, which make the readiness failing, and Kibana service never up.

It's here:
https://github.com/elastic/helm-charts/blob/11c23177c47b4f3ba1f0bcca59c9f3661e34d04e/kibana/templates/deployment.yaml#L94

When using this config:
```
          kibana.yml: |
            server.basePath: "/kibana"
            server.rewriteBasePath: true
```

We would need to have the readiness probe hitting `http '/kibana/app/kibana'`.

Thank you.

Crazybus:
I have opened #107 to make this path configurable.

Crazybus:
Also: Thank you for reporting this!

johnmarcou:
That is great! Thank you for reactivity :)

dkirrane:
Hi, just wondering what version chart this is in? Couldn't see it in `elastic/kibana` `7.0.0-alpha1`

Crazybus:
> Hi, just wondering what version chart this is in? Couldn't see it in elastic/kibana 7.0.0-alpha1

This change just missed the last release (merged in one day after). I'll be releasing the next version later today or tomorrow which will include this.

Crazybus:
@dkirrane this has now been released https://github.com/elastic/helm-charts/releases/tag/7.0.1-alpha1

PR #107, opened 2019-04-18
[kibana] Make health check path configurable

Crazybus:
Fixes: #103


dkirrane:
Hi, When just wondering when this will get added to the Helm chart?

PR #115, opened 2019-05-01
[elasticsearch] Add upgrade integration test

Crazybus:
This test will install the latest chart release and then attempt to
upgrade to what is currently in git. This is designed to test two
things.

1. That any changes do not happen to the statefulset that isn't allowed
by Kubernetes. An example of this happening is #94 where the change
was not caught as part of pull request testing but was luckily caught
during manual testing before the release.
2. That a rolling upgrade actually works and results in a healthy
cluster at the end. The test will always override the
"terminationGracePeriod" to make sure that a rolling upgrade is always
done even if there are no changes.

Kubectl and helm have also been updated to the latest releases. Oddly
enough there was a breaking change in the mock value used for release
names where it is now lowercased for some reason.

The kubectl upgraded was needed to get the handy timeout flag for the
`kubectl rollout status`command to make sure that jobs don't hang
forever if something does actually go wrong. This command is needed
because the `helm --wait` flag only works `apps/v1` resources. See
https://github.com/helm/helm/issues/3173 for more information.


Crazybus:
CI failure when downloading kubectl despite having 5 retries. jenkins test this please

```
curl: (7) Failed to connect to 2607:f8b0:4001:c15::80: Cannot assign requested address
The command '/bin/sh -c curl --retry 5 -O https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl &&     mv kubectl /usr/local/bin/ &&     chmod a+x /usr/local/bin/kubectl &&     kubectl version --client' returned a non-zero code: 7
make: *** [build] Error 7
```

Crazybus:
Jenkins test this please

PR #117, opened 2019-05-01
Filebeat Chart

tylerjl:

@Crazybus pushing what I have thus far. I think the actual body of the yaml templates, template tests, and docs are mostly complete, the remaining steps from my perspective are:

- [ ] Adding goss testing to check for end-to-end "are filebeat indices being created" tests, which will involve releasing ES to a test cluster
- [ ] Writing up accompanying CI definitions

I've pushed this to a branch on this repo instead of my fork so feel free to tinker with the branch if you want, I'm fine with either of us writing those tests and CI jobs, whoever gets to it first.

Implementation notes:
- I left out PDBs since they don't make sense for a DaemonSet but left configurable tolerations in place, because they apply for a DaemonSet anyway and were in my copypasta :grin: 
- Old versions of the Chart used `emptyDir`, while the newer reference that the beats team maintains prefers `hostPath`, which I've reflected here. See [the associated commit that made this change](https://github.com/elastic/beats/commit/3c30586a6068a61ea39147fd4c2b048499a27252#diff-ce74402547cad3c240e52c1e7190cd3a) for additional information.
- Do note that I'm using the [internal monitoring HTTP port](https://www.elastic.co/guide/en/beats/filebeat/7.0/http-endpoint.html) as a health check for the pod. I don't know whether I prefer this or using the config check `-c` flag, I leaned this way because it's a more definitive check in case the go process became unresponsive for whatever reason. The downside is that the docs mark this feature as experimental.
- This is the first chart that will potentially manage RBAC resources. I didn't do anything crazy here aside from permitting the chart to flip these off with a boolean `value`.


filebeat/templates/daemonset.yaml
```diff
@@ -0,0 +1,125 @@
...
+      {{- end }}
+      {{- if .Values.filebeatConfig }}
+      - name: filebeat-config
+        configMap:
+          defaultMode: 0600
+          name: {{ template "name" . }}-config
+      {{- end }}
+      - name: data
+        hostPath:
+          path: {{ .Values.hostPathRoot }}/{{ template "fullname" . }}-data
```
Crazybus: I think adding the namespace as part of the path would be a good idea too. Without it there will be collisions if there are multiple filebeat releases running in different namespaces but on the same physical host.

Crazybus: Release names need to be unique in helm but it is possible to have multiple helm/tiller installations in a cluster, and with helm 3 tiller will be gone so I suspect that release names will be namespaced too.

filebeat/templates/daemonset.yaml
```diff
@@ -0,0 +1,125 @@
...
+        - "-e"
+        - "-E"
+        - "http.enabled=true"
+        livenessProbe:
+{{ toYaml .Values.livenessProbe | indent 10 }}
+          httpGet:
+            port: monitor
+        ports:
+        - name: monitor
+          containerPort: 5066
```
Crazybus: Should probably make this configurable since it can be changed in filebeat. Not sure why but I'm sure someone has a valid usecase for wanting to customize it so may as well make it configurable.

Crazybus:
> Do note that I'm using the internal monitoring HTTP port as a health check for the pod. I don't know whether I prefer this or using the config check -c flag, I leaned this way because it's a more definitive check in case the go process became unresponsive for whatever reason. The downside is that the docs mark this feature as experimental.

This sounds good to me. I'm going to play with it a bit to see how it responds when the endpoint (such as elasticsearch) is down. It was only added in 6.4 but I don't see that being a problem as long as it is working in the latest 6.x and 7.x releases.

PR #121, opened 2019-05-03
Add integration tests and other tweaks for filebeat

Crazybus:
Apart from adding some integration tests and more templating tests I also made a few other changes:

* Added a default working filebeat config into the default values.yaml. This means the default install without any changes will result in logs being sent to the default elasticsearch chart. 
* Switched the livenessProbe to a curl since httpGet cannot check localhost. We could bind the monitoring API to a non-localhost port but this has security implications
* Added integration tests for the oss release of Filebeat/Kibana and Elasticsearch. I did this mostly because I wasn't 100% sure if the monitoring port was available in the oss release. And added it for the other charts to make sure features don't get introduced that would break the oss version.
* Added a readinessProbe that checks it can connect to its output. The livesnessProbe will restart filebeat if it isn't responding, and this will prevent upgrades from happening if the new configuration can't connect to its output. And also makes it easy for monitoring tools (like the data we get from metricbeat) to detect if a filebeat container is not healthy. 
* Fixed various name collisions. A few resources had a hardcoded "filebeat" name which would conflict if multiple filebeat instances were deployed to the namespace. When would this happen? One example is our integration testing, however I can also imagine blue/green deploy strategies or use cases where multiple filebeat daemonsets are deployed to the same cluster and namespace. 
* Added "namespace" to the hostPath that is being mounted. To make sure there are no conflicts if filebeat is deployed in two different namespaces with the same release name.


Crazybus:
>  it's probably worth a minor note in the filebeat chart's README

Agreed! I think adding this to a [Usage notes and getting started](https://github.com/elastic/helm-charts/tree/master/elasticsearch#usage-notes-and-getting-started) section like the Elasticsearch readme makes the most sense. 

Potentially giving an example of how to configure one of the other [many filebeat ouputs](https://www.elastic.co/guide/en/beats/filebeat/current/configuring-output.html) could also be useful. But hopefully an example of how to configure filebeat.yml for Elasticsearch and add the right environment variable should be enough. 

> I can PR that later though.

🙏

PR #123, opened 2019-05-03
[elasticsearch] make service configurable

kimxogus:

In case of [migration](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples/migration), master node name's prefix can be different. As we are already getting master node name only, I think checking if string is empty is enough


elasticsearch/templates/service.yaml
```diff
@@ -26,11 +34,10 @@ metadata:
     release: {{ .Release.Name | quote }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     app: "{{ template "uname" . }}"
-  annotations:
-    # Create endpoints also if the related pod isn't ready
-    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
```
kimxogus: annotations is deprecated https://github.com/kubernetes/kubernetes/pull/63742

kimxogus:
also added some service values

elasticsearch/templates/statefulset.yaml
```diff
@@ -242,7 +242,7 @@ spec:
           cleanup () {
             while true ; do
               local master="$(http "/_cat/master?h=node")"
-              if [[ $master == "{{ template "uname" . }}"* && $master != "${NODE_NAME}" ]]; then
+              if [[ $master && $master != "${NODE_NAME}" ]]; then
```
Crazybus: Why change this? It just seems less safe than what was already in there. Yes it will fail if there is an empty string, however if the API returns anything else (or some kind of weird error) then it is going to be a non-empty string and this will pass. 

Did you find an issue with the current logic that this fixes?

elasticsearch/templates/service.yaml
```diff
@@ -26,11 +34,10 @@ metadata:
...
 spec:
   clusterIP: None # This is needed for statefulset hostnames like elasticsearch-0 to resolve
+  # Create endpoints also if the related pod isn't ready
+  publishNotReadyAddresses: true
```
Crazybus: This was added in [1.8](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.8.md) so looks like this is the right time to switch 👍

elasticsearch/README.md
```diff
@@ -88,6 +88,9 @@ helm install --name elasticsearch elastic/elasticsearch --version 7.0.1-alpha1 -
...
+| `service.type`             | Type of elasticsearch service. [Service Types](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types)                                                                                                                                                                         | `ClusterIP`                                                                                                                    |
+| `service.annotations`      | Annotations that Kubernetes will use for the service. This will configure load balancer if `service.type` is `LoadBalancer` (https://kubernetes.io/docs/concepts/services-networking/service/#ssl-support-on-aws)                                                                                                          | `{}`                                                                                                                    |
```
Crazybus: Could you add an alias to the url to make it a bit shorter? Maybe adding it to [Annotations] would make the most sense

Crazybus:
jenkins test this please

elasticsearch/templates/statefulset.yaml
```diff
@@ -241,7 +241,7 @@ spec:
 
           cleanup () {
             while true ; do
-              local master="$(http "/_cat/master?h=node")"
+              local master="$(http "/_cat/master?h=node" || echo "")"
```
kimxogus: @Crazybus refactored here to make `$master` be empty string if API returns an error.

elasticsearch/templates/statefulset.yaml
kimxogus: Was my explanation not enough? Then I can restore this because this only happens in specific migration case.

Crazybus: > I think this is unnecessary because --fail option is passed to curl(https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/statefulset.yaml#L239), script will exit if an error is returned from API($master won't have a value).

Thanks for noticing this. This combination of `set -e` and `curl --fail` will mean that this script might exit early if the call fails. Ideally we want this loop to keep on running until we can see that a new master exists that isn't the current pod. Exiting early is not what we want to happen here. 

My concern is whether or not the API will return an error when there is no master or not. If the API returns a weird message like "no master yet" with a 200 then it's possible for the script to exit too early. Looking at the code or Elasticsearch it looks like it will actually [return a dash](https://github.com/elastic/elasticsearch/blob/044af2a4c4712bff4538b5eeee7b0f54dc2aba82/server/src/main/java/org/elasticsearch/rest/action/cat/RestMasterAction.java#L85-L90) (`-`) if there is no master found.

> Then I can restore this because this only happens in specific migration case.

Did it just mean that Kubernetes waited for the 120 second timeout while stopping each of the masters? 

I think what actually makes a lot more sense is to check that it starts with `{{ template "masterService" . }}`. That way this will always point to the prefix for the right master even during migrations like in https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/migration/README.md

> Was my explanation not enough?

Sorry for the slow reply. I'm currently on a work trip and only have a few quiet moments to sneak some work in. I'll be at Kubecon in Barcelona next week too so things will be slow for a little while. If you are at Kubecon come and say hi at the Elastic booth :)

kimxogus: > Did it just mean that Kubernetes waited for the 120 second timeout while stopping each of the masters?

Yes, I had to wait 120 seconds per each master node.

> I think what actually makes a lot more sense is to check that it starts with {{ template "masterService" . }}. That way this will always point to the prefix for the right master even during migrations like in /elasticsearch/examples/migration/README.md@master 

Changed to `{{ template "masterService" . }}`!

> I'm currently on a work trip and only have a few quiet moments to sneak some work in. I'll be at Kubecon in Barcelona next week too so things will be slow for a little while. If you are at Kubecon come and say hi at the Elastic booth :)

I'm not going to this Kubecon, but I hope to visit and say hi someday :)

elasticsearch/README.md
```diff
@@ -89,6 +89,9 @@ helm install --name elasticsearch elastic/elasticsearch --version 7.0.1-alpha1 -
...
+| `service.type`             | Type of elasticsearch service. [Service Types](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types)                                                                                                                                                                         | `ClusterIP`                                                                                                                    |
+| `service.annotations`      | Annotations that Kubernetes will use for the service. This will configure load balancer if `service.type` is `LoadBalancer` [Annotations](https://kubernetes.io/docs/concepts/services-networking/service/#ssl-support-on-aws)                                                                                             | `{}`                                                                                                                    |
+
```
Crazybus: This new line is causing the table formatting to break.

image

Crazybus:
jenkins test this please

elasticsearch/README.md
kimxogus: fixed

Crazybus:
jenkins test this please

Crazybus:
Thank you for another great contribution!

PR #127, opened 2019-05-09
Add metricbeat chart

Crazybus:
The first commit is a copy paste of the filebeat chart. If you want to see all of the changes after the `s/filebeat/metricbeat` you can look at: https://github.com/elastic/helm-charts/pull/127/files/f8aa4bfaa972ac72abb017acb484ffbf457295d2..8bc136528dfb40ed3fd563edf9322708e6873447

Couple of notes:

1. I don't really like the [add/remove `charts/` dance in .helmignore](https://github.com/elastic/helm-charts/pull/127/files#diff-a333719445f205838d2de7d8585c7735R6) however I had to add it because `helm lint --strict` will fail because of linting problems with kube-state-metrics child chart. And leaving it in `.helmignore` will mean that helm can't find the child chart :(
2. I made the decision for the metricbeat that uses kube-state-metrics to re-use the configmaps, secrets and extra envs from the daemonset. I did this because these secrets are always being used for the output to Elasticsearch. I couldn't imagine a situation where you would want to add a secret to one of them, but not the other. This simplifies the chart and removes a lot of duplication. The real reason we run a seperate deployment is so that we don't have each daemonset pod sending exactly the same information from the Kubernetes API. Since we theoretically could move the kube-state-metrics logic into the daemonset config (with the downside that you would have duplicated data and extra queries to the Kubernetes API) I can't see this causing any issues. 
3. I added in a couple of tests to make sure that the correct data really appears in Elasticsearch. This was to be sure that the daemonset metricbeat data and the deployment were making it in properly.


Crazybus:
Just realised the CI changes won't work in this PR since they are only deployed from master. I just deployed the changes from this branch to devops-ci.

jenkins test this please

Crazybus:
jenkins test this please

PR #135, opened 2019-05-16
[elasticsearch] add masterHeadlessService template

kimxogus:

In case of inter-namespace es cluster migration, `.Values.masterService` needs to contain full dns name like `es-master.ns.svc.cluster.local`.

As https://github.com/elastic/helm-charts/compare/master...kimxogus:elasticsearch/master-headless-service-helper?expand=1#diff-84f8d4f2bcce071fcd77bb865b4fc4a7L187 just append `-headless` to `{{ template "masterService" . }}`, discovery fails.

This PR adds a helper to remove `-headless` suffix if `.Values.masterService` contains `.` to fix this issue.

### DIFF
--- a/elasticsearch/templates/_helpers.tpl
+++ b/elasticsearch/templates/_helpers.tpl
@@ -27,6 +27,14 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 {{- end -}}
 
+{{- define "masterHeadlessService" -}}
+{{- if contains "." (include "masterService" .) -}}
+{{ template "masterService" . }}
+{{- else -}}
+{{ template "masterService" . }}-headless
+{{- end -}}
+{{- end -}}
+
 {{- define "endpoints" -}}
 {{- $replicas := .replicas | int }}
 {{- $uname := printf "%s-%s" .clusterName .nodeGroup }}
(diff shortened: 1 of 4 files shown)


kimxogus:
Also fixed missing `fullnameOverride` support

elasticsearch/templates/_helpers.tpl
```diff
@@ -27,6 +35,14 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- end -}}
 {{- end -}}
 
+{{- define "masterHeadlessService" -}}
+{{- if contains "." (include "masterService" .) -}}
```
Crazybus: This only covers use-cases where someone is migrating from a different namespace. In the [migration guide](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples/migration) the recommendation is to just manually set the environment variable.

https://github.com/elastic/helm-charts/blob/97f0004c50bdda7bee34069d2fe06f4eead96207/elasticsearch/examples/migration/master.yml#L9-L11

I think it would better to use this method as it works for all use cases, and doesn't add any extra complexity into the templating. Especially for something that only needs to exist temporarily and be done once for migrating users.

elasticsearch/templates/_helpers.tpl
```diff
@@ -11,9 +11,17 @@ Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "fullname" -}}
```
Crazybus: This template is actually not being used at all. So it could be removed instead.

PR #140, opened 2019-05-27
Remove fsGroup from container level security context

### LINKED ISSUE #136: Filebeat: failed validation with Helm 2.14 / kubernet 1.14
**Chart version:**
7.0.1-alpha1

**Kubernetes version:**
1.14.1

DigitalOcean

**Helm Version:**
2.14.0

Helm rejects the install with:
```
Error: validation failed: error validating "": error validating data: 
ValidationError(DaemonSet.spec.template.spec.containers[0].securityContext):
unknown field "fsGroup" in io.k8s.api.core.v1.SecurityContext
```

My understanding is that the `podSecurityContext` [should be only set](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) in `spec.securityContext`, not in `spec.containers[].securityContext` like it is here:
https://github.com/elastic/helm-charts/blob/master/filebeat/templates/daemonset.yaml#L108

Crazybus:
Fixes #136


Crazybus:
> Agreed that a container that typically needs root probably doesn't need this 👍

Just to be clear since I wrongly didn't include the context in the PR (it was only in the issue). But setting the fsGroup here wasn't actually doing anything. The latest version of helm now validates whether or not the manifest is valid. Previously it was just silently ignored and dropped from the configuration. The fsGroup setting needs to go in `spec.securityContext` on the pod level and not in `spec.containers[].securityContext`

PR #144, opened 2019-05-31
Fix for volumemounts indents and also added the serviceAccount feature from kibana as have a need for it

jamestutton:

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -101,7 +101,7 @@ spec:
             name: {{ template "uname" . }}-config
         {{- end }}
       {{- if .Values.extraVolumes }}
-{{ tpl .Values.extraVolumes . | indent 6 }}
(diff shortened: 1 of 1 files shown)


Crazybus:
Thanks for working on and fixing this! I'm planning to add a complex example to the [automated example tests](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples) to make sure we can test all of these sorts of features and be sure that they are configured properly. Right now the [integration tests](https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/default/test/goss.yaml) only focus on testing the features used by the default install.

acondrat:
Would really like to see it merged! @jamestutton this looks like a very simple and straightforward fix. Do you plan to submit it anytime soon?

Thanks!

jmlrt:
Hello @jamestutton,
Thank for your PR.
I'll close it as `extraVolumes` was fixed in #245 and Service Account was implemented in another way in #265.

PR #154, opened 2019-06-06
[elasticsearch] Keystore integration

Crazybus:

### DIFF
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -111,6 +111,7 @@ helm install --name elasticsearch elastic/elasticsearch --set imageTag=7.3.0
 | `schedulerName`               | Name of the [alternate scheduler](https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/#specify-schedulers-for-pods)                                                                                                                                                                          | `nil`                                                                                                                     |
 | `masterTerminationFix`        | A workaround needed for Elasticsearch < 7.2 to prevent master status being lost during restarts [#63](https://github.com/elastic/helm-charts/issues/63)                                                                                                                                                                    | `false`                                                                                                                   |
 | `lifecycle`                   | Allows you to add lifecycle configuration. See [values.yaml](./values.yaml) for an example of the formatting.                                                                                                                                                                                                              | `{}`                                                                                                                      |
+| `keystore`                    | Allows you map Kubernetes secrets into the keystore. See the [config example](/elasticsearch/examples/config/values.yaml) and [how to use the keystore](#how-to-use-the-keystore)                                                                                                                                          | `[]`                                                                                                                      |
 
 ## Try it out
 
@@ -171,18 +172,55 @@ There are a couple reasons we recommend this.
 
 #### How to use the keystore?
 
-1. Create a Kubernetes secret containing the [keystore](https://www.elastic.co/guide/en/elasticsearch/reference/current/secure-settings.html)
-   ```
-   $ kubectl create secret generic elasticsearch-keystore --from-file=./elasticsearch.keystore
-   ```
-2. Mount it into the container via `secretMounts`
-   ```
-   secretMounts:
-   - name: elasticsearch-keystore
-     secretName: elasticsearch-keystore
-     path: /usr/share/elasticsearch/config/elasticsearch.keystore
-     subPath: elasticsearch.keystore
-   ```
+
+##### Basic example
+
+Create the secret, the key name needs to be the keystore key path. In this example we will create a secret from a file and from a literal string.
+
+```
+kubectl create secret generic encryption_key --from-file=xpack.watcher.encryption_key=./watcher_encryption_key
+kubectl create secret generic slack_hook --from-literal=xpack.notification.slack.account.monitoring.secure_url='https://hooks.slack.com/services/asdasdasd/asdasdas/asdasd'
+```
+
+To add these secrets to the keystore:
+```
+keystore:
+  - secretName: encryption_key
+  - secretName: slack_hook
+```
+
+##### Multiple keys
+
+All keys in the secret will be added to the keystore. To create the previous example in one secret you could also do:
+
+```
+kubectl create secret generic keystore_secrets --from-file=xpack.watcher.encryption_key=./watcher_encryption_key --from-literal=xpack.notification.slack.account.monitoring.secure_url='https://hooks.slack.com/services/asdasdasd/asdasdas/asdasd'
+```
+
+```
+keystore:
+  - secretName: keystore_secrets
+```
+
+##### Custom paths and keys
+
+If you are using these secrets for other applications (besides the Elasticsearch keystore) then it is also possible to specify the keystore path and which keys you want to add. Everything specified under each `keystore` item will be passed through to the `volumeMounts` section for [mounting the secret](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets). In this example we will only add the `slack_hook` key from a secret that also has other keys. Our secret looks like this:
+
+```
+kubectl create secret generic slack_secrets --from-literal=slack_channel='#general' --from-literal=slack_hook='https://hooks.slack.com/services/asdasdasd/asdasdas/asdasd'
+```
+
+We only want to add the `slack_hook` key to the keystore at path `xpack.notification.slack.account.monitoring.secure_url`.
+
+```
+keystore:
+  - secretName: slack_secrets
+    items:
+    - key: slack_hook
+      path: xpack.notification.slack.account.monitoring.secure_url
+```
+
+You can also take a look at the [config example](/elasticsearch/examples/config/) which is used as part of the automated testing pipeline.
 
 #### How to enable snapshotting?
 
--- a/elasticsearch/examples/config/values.yaml
+++ b/elasticsearch/examples/config/values.yaml
@@ -1,0 +1,31 @@
+---
+
+clusterName: "config"
+replicas: 1
+
+extraEnvs:
+  - name: ELASTIC_PASSWORD
+    valueFrom:
+      secretKeyRef:
+        name: elastic-credentials
+        key: password
+  - name: ELASTIC_USERNAME
+    valueFrom:
+      secretKeyRef:
+        name: elastic-credentials
+        key: username
+
+# This is just a dummy file to make sure that
+# the keystore can be mounted at the same time
+# as a custom elasticsearch.yml
+esConfig:
+  elasticsearch.yml: |
+    path.data: /usr/share/elasticsearch/data
+
+keystore:
+  - secretName: elastic-config-secret
+  - secretName: elastic-config-slack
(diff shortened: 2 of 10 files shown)


elasticsearch/README.md
```diff
@@ -171,18 +172,9 @@ There are a couple reasons we recommend this.
...
-   ```
-2. Mount it into the container via `secretMounts`
-   ```
-   secretMounts:
-   - name: elasticsearch-keystore
-     secretName: elasticsearch-keystore
-     path: /usr/share/elasticsearch/config/elasticsearch.keystore
-     subPath: elasticsearch.keystore
-   ```
+Take a look a the [config example](/elasticsearch/examples/config/values.yaml) which has a tested version of adding strings and files to the keystore.
```
jordansissel: Might be helpful to have a note explicitly stating that any external values (secrets, etc) need to be explicitly set as env vars (which is what the example does). Thoughts?

elasticsearch/examples/config/values.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - name: elastic-config-secret
+    secretName: elastic-config-secret
+    path: /usr/share/elasticsearch/config/secret
+
+keystore:
+  enabled: true
+  strings:
+    bootstrap.password: '${ELASTIC_PASSWORD}'
+    xpack.notification.slack.account.monitoring.secure_url: '${SLACK_URL}'
+  files:
```
jordansissel: cloud-on-k8s doesn't have a specific concept of file vs string keystore entries: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-es-secure-settings.html

I don't recall why that is, but I wonder if we can have the same concepts for this on both the chart and the operator?

Crazybus: I remember a change was made (initiated by the cloud-on-k8s team) to have the Elasticsearch keystore treat all strings and files the same internally. 

My original assumption was that the operator would be doing something special to take these secrets, generate a keystore and create that as a secret. But their approach is also [fully init container based](https://github.com/elastic/cloud-on-k8s/blob/0f5aab786ea99f1374dceca1e86fca2c385de117/operators/pkg/controller/common/keystore/initcontainer.go#L33-L51). All secrets are mounted as files and then the init script loops over them and adds them to the keystore. 

I like this a lot more than my solution, mainly because you don't need to think about mapping the files and strings into the container yourself. And the other huge security advantage is that these secrets won't still exist as environment variables or mounts in the actual running Elasticsearch container. 

One downside to this approach is that it looks like the spec only allows a single secret. The downside of this being that you now need to fully recreate your keystore secret everytime you want to add something new. 

Making it a list would solve this so I'll ping the team and see what their thoughts on it are. 
```
spec:
  secureSettings:
    - secretName: your-secure-settings-secret
    - secretName: other-secure-settings-secret
```

I'm going to take a stab at implementing this in the same way.

Crazybus:
@jordansissel I have updated it to match how cloud-on-k8s is doing it (which is a much better approach). Issues have also been opened to implement the suggestions I had around allowing multiple secrets, and the ability to set all parts of the volumeMount spec to allow specifying a different path than the key of the secret.

https://github.com/elastic/cloud-on-k8s/issues/1457
https://github.com/elastic/cloud-on-k8s/issues/1458

Please take another look :)

elasticsearch/templates/statefulset.yaml
```diff
@@ -111,6 +111,14 @@ spec:
           configMap:
             name: {{ template "uname" . }}-config
         {{- end }}
+{{ if .Values.keystore }}
+        - name: keystore
+          emptyDir: {}
+{{ end }}
+        {{- range .Values.keystore }}
```
jordansissel: curious, why not have this `{{- range ...}}` inside the `{{ if ... ` block above? I I don't think there'd be any behavior changes, just curious about the style choice.

elasticsearch/templates/statefulset.yaml
```diff
@@ -111,6 +111,14 @@ spec:
           configMap:
             name: {{ template "uname" . }}-config
         {{- end }}
+{{ if .Values.keystore }}
```
jordansissel: Did you want to indent this and use `{{- if ...` ? Not sure if we're concerned with indentation style.

elasticsearch/templates/statefulset.yaml
```diff
@@ -129,6 +137,38 @@ spec:
...
+          elasticsearch-keystore create
+
+          for i in /tmp/keystoreSecrets/*/*; do
+            key=$(basename $i)
+            echo "Adding file $i to keystore key $key"
+            elasticsearch-keystore add-file "$key" "$i"
+          done
+
+          # Add the bootstrap password since otherwise the Elasticsearch entrypoint tries to do this on startup
+          [ ! -z "$ELASTIC_PASSWORD" ] && echo $ELASTIC_PASSWORD | elasticsearch-keystore add -x bootstrap.password
```
jordansissel: Needs quotation marks around the second var, `echo "$ELASTIC_PASSWORD"` otherwise a password with whitespace won't be set correctly.

jordansissel: It might be too complicated to add to the tests, but maybe we could add a test to cover this case (whitespace in a password, such as `ELASTIC_PASSWORD="hello       world     "`

Crazybus: Great catch! Fixed! I thought I had copy pasted this from the ELasticsearch docker file, but looks like my copy paste skills failed since it is properly quoted in there: https://github.com/elastic/elasticsearch/blob/31f6e78acdb8dfc7c042be4ad0bc0bdea003e055/distribution/docker/src/docker/bin/docker-entrypoint.sh#L88

elasticsearch/templates/statefulset.yaml
Crazybus: Fixed! It would be nice to add a "white space detection" check in the pull request testing and do a one off PR to get things back in line. Right now its a bit of a mess. It's a shame that `helm lint --strict` doesn't already cover this.

elasticsearch/templates/statefulset.yaml
Crazybus: This is just a leftover from the original refactoring. But you are right that it doesn't change anything. I fixed it up to make it as explicit as possible.

Crazybus:
Thanks for the review @jordansissel, all comments have been addressed. Please take another look :)

Crazybus:
:heart:

And thank you for the great suggestions and finding bugs!

dcvtruong:
@Crazybus What is the correct usage for "keystore: []" in values.yaml? I try several different format for '[]' format and it was not working.

keystore: 
   xpack.notification.slack.account.monitoring.secure_url: ''https://webhook_url"

Crazybus:
The simpliest example is:

https://github.com/elastic/helm-charts/blob/a49995539e4347e8729437cfbb54299023c81306/elasticsearch/examples/config/values.yaml#L25-L26

With the secret created like this:

https://github.com/elastic/helm-charts/blob/a49995539e4347e8729437cfbb54299023c81306/elasticsearch/examples/config/Makefile#L12

dcvtruong:
@Crazybus Thanks.

harsh4870:
i have tried adding GCS service account key in keystone but getting error

created kubernetes secret and added file into it and added secret into keystore

but pod are crashing also my i am using custom docker image with plugin installed but getting error like : 

```
csearch", "node.name": "elasticsearch-master-0",  "message": "uncaught exception in thread [main]" ,
"stacktrace": ["org.elasticsearch.bootstrap.StartupException: java.lang.IllegalArgumentException: unknown secure setting [gcs_backup_key.json] please check that any required plugins are installed, or check the breaking changes documentation for removed settings",
"at org.elasticsearch.bootstrap.Elasticsearch.init(Elasticsearch.java:163) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.bootstrap.Elasticsearch.execute(Elasticsearch.java:150) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:86) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.cli.Command.mainWithoutErrorHandling(Command.java:124) ~[elasticsearch-cli-7.3.2.jar:7.3.2]",
"at org.elasticsearch.cli.Command.main(Command.java:90) ~[elasticsearch-cli-7.3.2.jar:7.3.2]",
"at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:115) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:92) ~[elasticsearch-7.3.2.jar:7.3.2]",
"Caused by: java.lang.IllegalArgumentException: unknown secure setting [gcs_backup_key.json] please check that any required plugins are installed, or check the breaking changes documentation for removed settings",
"at org.elasticsearch.common.settings.AbstractScopedSettings.validate(AbstractScopedSettings.java:531) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.common.settings.AbstractScopedSettings.validate(AbstractScopedSettings.java:476) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.common.settings.AbstractScopedSettings.validate(AbstractScopedSettings.java:447) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.common.settings.AbstractScopedSettings.validate(AbstractScopedSettings.java:418) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.common.settings.SettingsModule.<init>(SettingsModule.java:149) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.node.Node.<init>(Node.java:357) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.node.Node.<init>(Node.java:258) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.bootstrap.Bootstrap$5.<init>(Bootstrap.java:221) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.bootstrap.Bootstrap.setup(Bootstrap.java:221) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.bootstrap.Bootstrap.init(Bootstrap.java:349) ~[elasticsearch-7.3.2.jar:7.3.2]",
"at org.elasticsearch.bootstrap.Elasticsearch.init(Elasticsearch.java:159) ~[elasticsearch-7.3.2.jar:7.3.2]",
"... 6 more"] }
```
created secret using 

`kubectl create secret generic gcs-backup-key --from-file=gcs_backup_key.json=gcs_backup_key.json`

i want to set cronjob for setting regular backup let me know if missing anything thanks in advance for help.

PR #158, opened 2019-06-11
Update beta notice and add chart descriptions

Crazybus:


tetianakravchenko:
should it also be adjusted in the main README file: https://github.com/elastic/helm-charts/blob/master/README.md#elastic-stack-kubernetes-helm-charts ?

PR #169, opened 2019-06-17
Add GKE 1.13 to automated testing suite

Crazybus:
The only change needed here was to not use goss to directly check if the port is open. With the default install of GKE 1.13 the containers are now being detected as listening on ipv4, when on previous versions it is on ipv6. In reality we only care that it is listening publicly so that services work. So instead the tests have been updated to test this rather than directly looking at the port. This also makes it more likely for the tests to work properly on other Kubernetes providers in the future.


Crazybus:
jenkins test this please

Crazybus:
Sorry I hid the in-depth reasoning in one of the git commits: https://github.com/elastic/helm-charts/pull/169/commits/1292370505e18ff286c230a912089c3c62ffe4e8.

The issue with goss right now is that it can't properly handle situations where a port is listening on tcp4 and tcp6 addresses https://github.com/aelsabbahy/goss/issues/149. The way that it is being detected means that it only works for one. So even though making a http request to `0.0.0.0` works, checking `tcp4:0.0.0.0` is open will fail.

But your point is still valid for an ipv6 only environment. I think the most portable option would be to.

1. Hit localhost when testing the the `/` endpoint.
2. Do a test against the service for any cluster related tests (checking health green)

That way ipv4/ipv6 is being left out of it altogether. PR incoming with these changes.

Issue #166, opened 2019-06-17
Error from deploying elasticsearch + security


bamb00:
**Chart version:**
elasticsearch-7.1.1
**Kubernetes version:**
1.13.0
**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
kubespray
**Helm Version:**
v2.13.0
**`helm get release` output**

e.g. `helm get elasticsearch` (replace `elasticsearch` with the name of your helm release)

```
```

**Describe the bug:**
Run the 'make' from the security directory.

**Steps to reproduce:**

1. cd to the examples/security dir
2. type 'make', hit enter 
3. install failed with "Error: release helm-es-security failed: timed out waiting for the condition" and "make: *** [install] Error 1"

**Expected behavior:**

**Provide logs and/or server output (if relevant):**

		$ make
		kubectl delete secrets elastic-credentials elastic-certificates elastic-certificate-pem || true && \
		vault read -field=value secret/devops-ci/helm-charts/elasticsearch/security/certificates | base64 --decode > elastic-certificates.p12 && \
		vault read -field=value secret/devops-ci/helm-charts/elasticsearch/security/certificate-pem | base64 --decode > elastic-certificate.pem && \
		kubectl create secret generic elastic-credentials  --from-literal=password=changeme --from-literal=username=elastic && \
		kubectl create secret generic elastic-certificates --from-file=elastic-certificates.p12 && \
		kubectl create secret generic elastic-certificate-pem --from-file=elastic-certificate.pem && \
		rm -f elastic-certificates.p12 elastic-certificate.pem
		Error from server (NotFound): secrets "elastic-credentials" not found
		Error from server (NotFound): secrets "elastic-certificates" not found
		Error from server (NotFound): secrets "elastic-certificate-pem" not found
		/bin/sh: line 1: vault: command not found
		/bin/sh: line 2: vault: command not found
		secret/elastic-credentials created
		secret/elastic-certificates created
		secret/elastic-certificate-pem created
		helm upgrade --wait --timeout=600 --install --values ./security.yml helm-es-security ../../ ; \

		Release "helm-es-security" does not exist. Installing it now.
		Error: release helm-es-security failed: timed out waiting for the condition
		make: *** [install] Error 1

**Any additional context:**

Crazybus:
Sorry it looks like these docs got a bit out of sync with the setup. I'm working on a PR now to update the docs. The two issues are:

[...]

PR #170, opened 2019-06-17
[elasticsearch] Update security example docs to match reality

Crazybus:
Over time the security example changed a bit and these instructions were
no longer valid or working.

Fixes: #166

### DIFF
--- a/elasticsearch/README.md
+++ b/elasticsearch/README.md
@@ -129,33 +129,29 @@ make
 
 ### Security
 
-A cluster with X-Pack security enabled
+A cluster with security enabled
 
-* Generate SSL certificates following the [official docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.7/configuring-tls.html#node-certificates)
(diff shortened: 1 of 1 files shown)


elasticsearch/README.md
```diff
@@ -129,33 +129,29 @@ make
...
   ```
   cd examples/security
-  make
+  helm upgrade --wait --timeout=600 --install --values ./security.yml elasticsearch ../../
   ```
 * Attach into one of the containers
 
   ```
-  kubectl exec -ti $(kubectl get pods -l release=helm-es-security -o name | awk -F'/' '{ print $NF }' | head -n 1) bash
+  kubectl exec -ti $(kubectl get --no-headers=true pods -l release=elasticsearch -o custom-columns=:metadata.name | head -n 1 ) bash
```
tylerjl: ` `  :scream:

Crazybus: One step closer to being able to do basic operations without needing third party tooling!

PR #171, opened 2019-06-17
[elasticsearch] Run as 1000

naseemkullah:


naseemkullah:
CLA has been signed

elasticsearch/templates/statefulset.yaml
```diff
@@ -124,6 +124,9 @@ spec:
       {{- end }}
       containers:
       - name: "{{ template "name" . }}"
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 1000
```
Crazybus: This value will need to be configurable for systems where the running user is not always 1000 (like Openshift). I think it makes the most sense to expose this variable in the values.yaml with these as the defaults.

Crazybus:
jenkins test this please

elasticsearch/templates/statefulset.yaml
naseemkullah: Thanks, I took a similar approach as `kibana` with `podSecurityContext` and `securityContext` in the values file.

elasticsearch/templates/statefulset.yaml
```diff
@@ -48,8 +48,10 @@ spec:
       {{- if .Values.schedulerName }}
       schedulerName: "{{ .Values.schedulerName }}"
       {{- end }}
+{{- if .Values.podSecurityContext }}
       securityContext:
-        fsGroup: {{ .Values.fsGroup }}
```
Crazybus: Removing this is a backwards incompatible change. We can deprecate it, however anyone who already has it set should still be able to upgrade. 

My suggestion would be:

1. Add it back in underneath the `{{ toYaml .Values.podSecurityContext | indent 8 }}` that has been added below.  
2. Set fsGroup to an empty value in values.yaml
3. Leave it in the readme but mark it as a deprecated option. 

This will keep the change backwards compatible for anyone who has it set to a custom value.

elasticsearch/values.yaml
```diff
@@ -129,8 +129,16 @@ updateStrategy: RollingUpdate
 # of your pods to be unavailable during maintenance
 maxUnavailable: 1
 
- # GroupID for the elasticsearch user. The official elastic docker images always have the id of 1000
-fsGroup: 1000
+podSecurityContext:
+  fsGroup: 2000
```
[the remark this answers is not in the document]
naseemkullah: Just thought I'd use a number that does not map to an actual group name (still works) but let's go with 1000 instead! Thanks

elasticsearch/templates/statefulset.yaml
naseemkullah: Changes applied! Not sure what happens if both values are set however.

elasticsearch/values.yaml
```diff
@@ -129,8 +129,20 @@ updateStrategy: RollingUpdate
...
 maxUnavailable: 1
 
- # GroupID for the elasticsearch user. The official elastic docker images always have the id of 1000
-fsGroup: 1000
+podSecurityContext:
+  fsGroup: 1000
+
+# The following value is deprecated,
+# please use the above podSecurityContext.fsGroup instead
+fsGroup: {}
```
Crazybus: This should be an empty value and not an empty hash I think.

elasticsearch/templates/statefulset.yaml
```diff
@@ -49,7 +49,8 @@ spec:
       schedulerName: "{{ .Values.schedulerName }}"
       {{- end }}
       securityContext:
-        fsGroup: {{ .Values.fsGroup }}
+{{ toYaml .Values.podSecurityContext | indent 8 }}
+        fsGroup: {{ .Values.fsGroup }} # Deprecated value, please use .Values.podSecurityContext.fsGroup
```
Crazybus: You will need to wrap this with a `{{ if .Values.fsGroup }}`. Its fine to have the same field twice as the second one is chosen. However this will always set fsGroup to `.Values.fsGroup` when we only want to do it when it is actually defined/set.

elasticsearch/values.yaml
naseemkullah: Set to `null`!

elasticsearch/templates/statefulset.yaml
naseemkullah: Oh good point! Changed.

Crazybus:
jenkins test this please

Crazybus:
CI failure is coming from the linting test:

```
09:27:40 helm lint --strict ./
09:27:40 ==> Linting ./
09:27:40 [ERROR] templates/: render error in "elasticsearch/templates/statefulset.yaml": template: elasticsearch/templates/statefulset.yaml:53:22: executing "elasticsearch/templates/statefulset.yaml" at <.Values.fsGroup>: map has no entry for key "fsGroup"
```

elasticsearch/values.yaml
```diff
@@ -129,8 +129,20 @@ updateStrategy: RollingUpdate
...
 maxUnavailable: 1
 
- # GroupID for the elasticsearch user. The official elastic docker images always have the id of 1000
-fsGroup: 1000
+podSecurityContext:
+  fsGroup: 1000
+
+# The following value is deprecated,
+# please use the above podSecurityContext.fsGroup instead
+fsGroup: null
```
Crazybus: I think setting it to an empty string will make the linter happy:

```
fsGroup: ''
```

naseemkullah: Ah right, thanks!

Crazybus:
jenkins test this please

naseemkullah:
Seems like Jenkins nodes are offline? :)
```
Configuration elastic / helm-charts - pull-request - cluster creation » 1.12,docker&&virtual is still in the queue: All nodes of label ‘docker&&virtual’ are offline
```

Crazybus:
> Seems like Jenkins nodes are offline? :)

This is expected, a fresh VM is spun up for every build and is set offline so that it can't be re-used. The build failed because creating the test GKE cluster failed. 

jenkins test this please

naseemkullah:
My pleasure!

PR #172, opened 2019-06-17
[kibana] Run as 1000

naseemkullah:


kibana/templates/deployment.yaml
```diff
@@ -63,6 +63,9 @@ spec:
     {{- end }}
       containers:
       - name: kibana
+        securityContext:
```
Crazybus: Same comment as the Elasticsearch PR. It makes the most sense to expose the securityContext to be configurable just like with the pod level securityContext.

https://github.com/elastic/helm-charts/blob/26fe7570fb668b909d11cddb6c5fa2f8cfaed288/kibana/templates/deployment.yaml#L30-L33

naseemkullah: Sounds good! I've added it as a configurable var. I'll go ahead and do it the same in ES to keep it consistent.

naseemkullah:
I took the liberty of adding extra `securityContext` as it is a best practice. I was unable to use a RO filesystem however, even when mounting an `emptyDir` volume on `/usr/share/kibana/optimize`.

kibana/values.yaml
```diff
@@ -44,8 +44,17 @@ kibanaConfig: {}
 #       nestedkey: value
 
 # If Pod Security Policy in use it may be required to specify security context as well as service account
-podSecurityContext: {}
-  #runAsUser: "place the user id here"
+
+podSecurityContext:
+  fsGroup: 2000
```
Crazybus: What is group id 2000? It doesn't exist in the kibana docker image. The kibana group has id 1000.

Crazybus:
jenkins test this please

kibana/values.yaml
naseemkullah: Thought I'd use a number that does not map to an actual group (still works) but yes let's go with 1000 that maps to an existing group instead!

Crazybus:
jenkins test this please

Crazybus:
Thank you so much for adding this in! 🔐

naseemkullah:
My pleasure!

PR #173, opened 2019-06-17
[kibana] Keep in sync with current standards

naseemkullah:

### DIFF
--- a/kibana/templates/configmap.yaml
+++ b/kibana/templates/configmap.yaml
@@ -1,12 +1,10 @@
 {{- if .Values.kibanaConfig }}
----
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ template "fullname" . }}-config
+  name: {{ include "kibana.fullname" . }}-config
   labels:
-    app: {{ .Chart.Name }}
-    release: {{ .Release.Name | quote }}
+{{ include "kibana.labels" . | indent 4 }}
 data:
 {{- range $path, $config := .Values.kibanaConfig }}
   {{ $path }}: |
--- a/kibana/templates/ingress.yaml
+++ b/kibana/templates/ingress.yaml
@@ -1,38 +1,36 @@
 {{- if .Values.ingress.enabled -}}
-{{- $fullName := include "fullname" . -}}
-{{- $servicePort := .Values.service.port -}}
-{{- $ingressPath := .Values.ingress.path -}}
+{{- $fullName := include "kibana.fullname" . -}}
 apiVersion: extensions/v1beta1
 kind: Ingress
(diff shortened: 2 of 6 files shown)


kibana/templates/configmap.yaml
```diff
@@ -1,12 +1,10 @@
 {{- if .Values.kibanaConfig }}
----
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ template "fullname" . }}-config
+  name: {{ include "kibana.fullname" . }}-config
   labels:
-    app: {{ .Chart.Name }}
-    release: {{ .Release.Name | quote }}
```
Crazybus: Removing these labels is a backwards incompatible change. Adding new labels is fine but there is no need to remove these ones.

kibana/templates/ingress.yaml
```diff
@@ -1,38 +1,36 @@
...
   {{- end }}
 {{- end }}
   rules:
   {{- range .Values.ingress.hosts }}
-    - host: {{ . }}
+    - host: {{ .host | quote }}
       http:
         paths:
-          - path: {{ $ingressPath }}
+        {{- range .paths }}
```
Crazybus: This breaks backwards compatibility. Whats the use case for wanting multiple paths for the ingress record? If there is a good one we can look at ways to add this in a backwards compatible way.

Crazybus:
jenkins test this please

kibana/templates/ingress.yaml
naseemkullah: You are right!

kibana/templates/configmap.yaml
naseemkullah: I think these charts are at an early enough stage that we can slip in a few backwards incompatible changes ;) But your call!

Crazybus: The reason the charts are following the stack versioning is so that users no they can upgrade without breaking changes. If there is a really good reason to making a breaking change (with a documented way to preserve backwards compatibly) then I would be ok with it. However it would need to have a serious security or availability impact. Removing some labels doesn't really fall into this category. 

I know of a lot of people using labels for filtering and alerting of logs. Someone could upgrade their cluster, things would still work, but suddenly have broken alerting or dashboards because the labels were removed. 

Another even better example is the automated testing done in this repo uses the `release` label. So this change will fail in CI because it can't find which pod to use for testing:

https://github.com/elastic/helm-charts/blob/9dd7bbe2d6bfb113bd63d313063f05c3a024f4dd/helpers/examples.mk#L4

naseemkullah:
More pain than gain here as well. Closing.

Issue #150, opened 2019-06-05
Elasticsearch: Resources created by chart are keyed off of clusterName rather than Release.Name


davestephens:
**Chart version:** 7.1.0

**Kubernetes version:** 1.14

**Kubernetes provider:** Azure

**Helm Version:** 2.14

**Describe the bug:**
The resources created by the chart are keyed off of the `clusterName` value, rather than `Release.Name`. 

This doesn't follow normal Helm patterns, meaning that if you run `helm install elastic/elasticsearch` twice (with different release names but not specifying a `clusterName`), the second deployment will fail, as well as break the first one. 

**Steps to reproduce:**

1. `helm install --name dude-elasticsearch elastic/elasticsearch`
2. Wait for cluster to come up.
3. `helm install --name sweet-elasticsearch elastic/elasticsearch`
4. Observe brokenness of dude-elasticsearch, and failure of deployment of `sweet-elasticsearch`.

**Expected behavior:**
It should be possible to run the chart twice without having to specify a `clusterName` and end up with two working Elasticsearch clusters within k8s.

Crazybus:
> The resources created by the chart are keyed off of the clusterName value, rather than Release.Name.
> This doesn't follow normal Helm patterns, meaning that if you run helm install elastic/elasticsearch twice (with different release names but not specifying a clusterName), the second deployment will fail, as well as break the first one.

When I was developing the chart I was originally using the `Release.Name`. I've looked back through the old commit history (before it was first publicly released) and I found the commit message where it was switched which has the reasoning the it was to make it possible to deploy clusters with [multiple node groups](https://github.com/elastic/helm-charts/blob/master/elasticsearch/README.md#multi). Which is where groups of nodes are part of the same Elasticsearch cluster but deployed using separate helm releases. This makes it possible to replace groups of nodes (you can't do resizing of pv for all Kubernetes clusters yet) and to make up whatever architecture you want. 

However, past me didn't go into detail about why this was needed. Looking at the chart now I can't figure out why it couldn't just use `Release.Name` for all the resources instead. So I have the exact same question as you now!

So I wish I had a good answer, but it seems like either past me made a bad decision, or its possible that trying to switch it back to `Release.Name` and setup a multi node cluster might help to explain it. 

[...]

PR #174, opened 2019-06-18
[elasticsearch] Keep in sync with current standards

naseemkullah:

### DIFF
--- a/elasticsearch/templates/ingress.yaml
+++ b/elasticsearch/templates/ingress.yaml
@@ -1,5 +1,7 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "elasticsearch.fullname" . -}}
+{{- $servicePort := .Values.httpPort -}}
+{{- $ingressPath := .Values.ingress.path -}}
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
@@ -26,11 +28,9 @@ spec:
     - host: {{ .host | quote }}
       http:
         paths:
-        {{- range .paths }}
-          - path: {{ . }}
+          - path: {{ $ingressPath }}
             backend:
               serviceName: {{ $fullName }}
-              servicePort: http
-        {{- end }}
+              servicePort: {{ $servicePort }}
   {{- end }}
 {{- end }}
--- a/elasticsearch/templates/ingress.yaml
+++ b/elasticsearch/templates/ingress.yaml
@@ -1,38 +1,36 @@
 {{- if .Values.ingress.enabled -}}
-{{- $fullName := include "uname" . -}}
-{{- $servicePort := .Values.httpPort -}}
-{{- $ingressPath := .Values.ingress.path -}}
+{{- $fullName := include "elasticsearch.fullname" . -}}
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
   labels:
-    app: {{ .Chart.Name }}
-    release: {{ .Release.Name }}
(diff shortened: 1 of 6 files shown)


elasticsearch/templates/ingress.yaml
```diff
@@ -1,38 +1,36 @@
...
-{{- $fullName := include "uname" . -}}
-{{- $servicePort := .Values.httpPort -}}
-{{- $ingressPath := .Values.ingress.path -}}
+{{- $fullName := include "elasticsearch.fullname" . -}}
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
   labels:
-    app: {{ .Chart.Name }}
```
Crazybus: Removing these labels breaks backwards compatibility.

elasticsearch/templates/ingress.yaml
```diff
@@ -1,38 +1,36 @@
...
 {{- end }}
   rules:
   {{- range .Values.ingress.hosts }}
-    - host: {{ . }}
+    - host: {{ .host | quote }}
       http:
         paths:
-          - path: {{ $ingressPath }}
+        {{- range .paths }}
+          - path: {{ . }}
```
Crazybus: Same question as for your Kibana PR. Is there a valid usecase for this? There is only one backend so I can't imagine why you would want certain paths going to different service ports.

elasticsearch/templates/statefulset.yaml
```diff
@@ -2,25 +2,23 @@
 apiVersion: apps/v1beta1
 kind: StatefulSet
 metadata:
-  name: {{ template "uname" . }}
+  name: {{ include "elasticsearch.fullname" . }}
```
Crazybus: This is a very big backwards compatibility change. Changing this will mean that everyone needs to re-deploy their clusters. We would need a very good reason to have to do this outside of a major version change. More discussion around this is happening in https://github.com/elastic/helm-charts/issues/150

naseemkullah: Let me know how you think we should proceed. I don't mind closing these PRs as they are mostly stylistic in nature.

elasticsearch/templates/ingress.yaml
naseemkullah: Agreed.

elasticsearch/templates/statefulset.yaml
naseemkullah: But they do fall in line with what you get now when you run `helm create`... so your call!

elasticsearch/templates/ingress.yaml
naseemkullah: Short term pain, long term gain? ;) Your call!

Crazybus: Could you explain what the gain is? Just sounds like a lot of pain to me. We can add the new labels without removing the old ones which is a win win.

elasticsearch/templates/statefulset.yaml
Crazybus: We are having the same conversations in two PRs :P So I'll just link to my reply from the Kibana PR. https://github.com/elastic/helm-charts/pull/173#discussion_r296128610

> But they do fall in line with what you get now when you run helm create... so your call!

We can consider this for the 8.0 release, but falling in line with the  `helm create` template is not a good enough reason to ask everybody to do a full recreate of all their clusters.

naseemkullah:
You are right, this brings more pain than gain then. Closing!

PR #187, opened 2019-06-24
Fix naming of headless node

timmjd:

Fixed incorect template parameters. Documentation ([# This should be set to clusterName + "-" + nodeGroup for your master group](https://github.com/elastic/helm-charts/blob/master/elasticsearch/values.yaml#L6)) did not match implementation.
* Service name depend on __uname__: [name: {{ template "uname" . }}-headless](https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/service.yaml#L34)
* Discover name depend on __masterService__: [value: "{{ template "masterService" . }}-headless"](https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/statefulset.yaml#L187)

values.yaml
```yaml
clusterName: "elasticsearch"
nodeGroup: "xxx"
```

Lead to error in startup:
```{"type": "server", "timestamp": "2019-06-24T21:05:19,992+0000", "level": "WARN", "component": "o.e.d.SeedHostsResolver", "cluster.name": "elasticsearch", "node.name": "elasticsearch-xxx-0",  "message": "failed to resolve host [elasticsearch-master-headless]" ...```

### DIFF
--- a/elasticsearch/templates/_helpers.tpl
+++ b/elasticsearch/templates/_helpers.tpl
@@ -21,7 +21,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 
 {{- define "masterService" -}}
 {{- if empty .Values.masterService -}}
-{{ .Values.clusterName }}-master
+{{ .Values.clusterName }}-{{ .Values.nodeGroup }}
 {{- else -}}
 {{ .Values.masterService }}
 {{- end -}}

### CODE REVIEW

File: elasticsearch/templates/_helpers.tpl [hunk 1]
```diff
@@ -21,7 +21,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 
 {{- define "masterService" -}}
 {{- if empty .Values.masterService -}}
-{{ .Values.clusterName }}-master
+{{ .Values.clusterName }}-{{ .Values.nodeGroup }}
```
Crazybus: This would be a breaking change for anyone running clusters with multiple nodes types and using the default settings. One example is our [multi example](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples/multi). So if you are having a problem trying to setup a multi node cluster I would recommend opening an issue with the problems you are running into. 


How and why this is works is explained in more detail in [clustering-and-node-discovery](https://github.com/elastic/helm-charts/tree/master/elasticsearch#clustering-and-node-discovery) in the readme.

timmjd: Hello @Crazybus, the example only does work due to `nodeGroup: "master"` inside the multi example.

If you change this to `nodeGroup: "foobar"`, it will stop working due to ElasticSearch will will search for a `discovery.zen.ping.unicast.hosts = multi-master` and not `discovery.zen.ping.unicast.hosts = multi-foobar`

Crazybus: Correct. As documented in the readme you also need to set `masterService` if you are changing the default `nodeGroup` for your group of masters.

> masterService | Optional. The service name used to connect to the masters. You only need to set this if your master nodeGroup is set to something other than master. See Clustering and Node Discovery for more information.

And also in the clustering and node discovery section that I linked to you above:

> If your group of master nodes has the default nodeGroup: master then you can just add new groups of nodes with a different nodeGroup and they will automatically discover the correct master. If your master nodes have a different nodeGroup name then you will need to set masterService to $clusterName-$masterNodeGroup.

If the documentation is unclear then we want to improve the documention. You proposed change would be a breaking change for everyone running multi node clusters with the default settings.

timmjd: Damn, you are right. Ok. Thanks for the clarification!

Issue #190, opened 2019-06-28
[elasticsearch] Add lifecycle hooks


tetianakravchenko:
**Describe the feature:**
Add option to provide custom start/stop hooks

**Describe a specific use case for the feature:**
For coordination nodes, I want to add pre-stop hook to wait before sending SIGTERM to complete already open connections. Otherwise, during the rolling restart client pods will be killed and all active connection will be dropped.

Crazybus:
>  Add option to provide custom start/stop hooks

👍 

> For coordination nodes, I want to add pre-stop hook to wait before sending SIGTERM to complete already open connections. Otherwise, during the rolling restart client pods will be killed and all active connection will be dropped.

Is Elasticsearch really not waiting for these connections to be completed? If the connections are really long running then it is possible you might need to increase the `terminationGracePeriod` to make sure that Kubernetes doesn't kill the pod too early. 

https://github.com/elastic/helm-charts/blob/2d309cbfb25ca0f5d3c3633445144d2c5ed1de54/elasticsearch/values.yaml#L135-L136

What would you be adding as a `preStop` hook to prevent this behavior? I'm interested since it sounds like something that maybe should be added into the helm chart by default. Or that it needs to also be fixed in Elasticsearch itself too.

tetianakravchenko:
@Crazybus thank you for your reply!

> Is Elasticsearch really not waiting for these connections to be completed?

I've tried with sample data provided here https://www.elastic.co/guide/en/kibana/current/tutorial-load-dataset.html (logstash indexes).
Setup:
3 nodes with dedicated client nodes:
```
test-client-0   1/1     Running       0          2m45s
test-client-1   1/1     Running       0          2m45s
test-master-0   1/1     Running   0          5m41s
test-master-1   1/1     Running   0          5h23m
test-master-2   1/1     Running   0          11m
```
IP address of `test-client-0`:
```
[root@test-client-0 elasticsearch]# hostname -i
10.32.22.115
```
I've used this query (query is not so important here), run it from `test-master-2` container:
```
[root@test-master-2 elasticsearch]# curl -XGET -o /dev/null  -H 'Content-Type: application/json' http://10.32.22.115:9200/_search -d'
{
 "query": {
    "bool": {
      "must": {
        "match_all": {}
      },
      "should" : [
        { "regexp" : { "ip" : ".<1-99>*<1-5><0-7>*" } },
        { "regexp" : { "agent" : ".oz*5*" } },
[...]

PR #197, opened 2019-07-02
[elasticsearch] Add option to provide custom start/stop hooks

tetianakravchenko:

PR should fix issue: https://github.com/elastic/helm-charts/issues/190


elasticsearch/templates/statefulset.yaml
```diff
@@ -215,10 +215,17 @@ spec:
...
           {{- range $path, $config := .Values.esConfig }}
-          - name: esconfig
+          - name: config
             mountPath: /usr/share/elasticsearch/config/{{ $path }}
             subPath: {{ $path }}
           {{- end -}}
+  {{- if .Values.hooks }}
+          {{- range $path, $config := .Values.hooks }}
+          - name: config
+            mountPath: /{{ $path }}
```
Crazybus: This is just identical to the existing block above it, with the only difference that is is being mounted into the root directory instead of the elasticsearch config directory. Would you have any issue with just mounting the files via `esConfig`? Just removes the code duplication that does exactly the same thing but in a different directory.

elasticsearch/templates/statefulset.yaml
```diff
@@ -276,3 +283,7 @@ spec:
         {{- end }}
       {{- end }}
       {{- end }}
+{{- if .Values.lifecycle }}
```
Crazybus: Could you add a test for this please. It will look something like:

```
def test_lifecycle_hooks():
    config = ''
    r = helm_template(config)
    c = r['statefulset'][uname]['spec']['template']['spec']['containers'][0]
    assert 'lifecycle' not in c

    config = '''
    lifecycle:
      preStop:
        exec:
          command: ["/bin/bash","/preStop"]
    '''
    r = helm_template(config)
    c = r['statefulset'][uname]['spec']['template']['spec']['containers'][0]

    assert c['lifecycle']['preStop']['exec']['command'] == ["/bin/bash","/preStop"]
```

elasticsearch/values.yaml
```diff
@@ -184,3 +192,11 @@ fullnameOverride: ""
 
 # https://github.com/elastic/helm-charts/issues/63
 masterTerminationFix: false
+
+lifecycle: {}
+  # preStop:
+  #   exec:
+  #     command: ["/bin/bash","/preStop"]
+  # preStop:
```
Crazybus: This example is the same as above. I think this one was meant to be `postStart`.

elasticsearch/templates/statefulset.yaml
tetianakravchenko: @Crazybus here is used different variables in `range`, do you suggest smth like:
```
{{- range $path, $config := .Values.esConfig }}
- name: config
   mountPath: /usr/share/elasticsearch/config/{{ $path }}
  subPath: {{ $path }}
{{- end -}}
{{- range $path, $config := .Values.hooks }}
- name: config
   mountPath: /usr/share/elasticsearch/config/{{ $path }}
    subPath: {{ $path }}
{{- end -}}
```
but anyhow here will be duplication. or?

elasticsearch/values.yaml
tetianakravchenko: done https://github.com/elastic/helm-charts/pull/197/commits/db535e2ae6c6e4d7642ab395b79f65779ed8935f, I've also moved `hooks` value closer to the `lifecycle` to make it more clear where should be defined content of scripts

elasticsearch/templates/statefulset.yaml
Crazybus: The same thing could be said about putting this under `hooks`. This is actually adding functionality for mounting a file into the container (the same thing that `esConfig` does) and has nothing to do with hooks. If we need a generic way to mount files into the container that are outside `esConfig` then a new more generic variable should be added to do that like `files. 

A hook specific one is also not even needed for this use case, instead you can just put the script inline directly in the lifecycle section which is what is documented in the Kubernetes examples:

```
lifecycle:
  postStart:
    exec:
      command: ["/bin/sh", "-c", "echo Hello from the postStart handler > /usr/share/message"]
```

And if you wanted to do this with a multiline script it would look like:

```
lifecycle:
  postStart:
    exec:
      command:
        - sh
        - -c
        - |
          #!/bin/bash
          set -euo pipefail
          echo Hello from the postStart handler > /usr/share/message
          echo this is a multiline script!
```

tetianakravchenko: @Crazybus thank you for the clarification! it totally makes sense. I've removed `hooks`, kept only `lifecycle` related changes. Could you please have a second look?

Crazybus:
jenkins test this please

tetianakravchenko:
@Crazybus there is some issue on k8s 1.13, not related to the PR :
```11:10:31 The command '/bin/sh -c yum -y install     https://packages.cloud.google.com/yum/pool/654962a43cb6bc6b47bc312424410b7c203e1beb73386d868d9e614839c82609-google-cloud-sdk-215.0.0-1.el7.noarch.rpm &&     gcloud config set core/disable_usage_reporting true &&     gcloud config set component_manager/disable_update_check true &&     gcloud version' returned a non-zero code: 1
11:10:46 Unable to find image 'helm-charts:latest' locally
11:10:46 docker: Error response from daemon: pull access denied for helm-charts, repository does not exist or may require 'docker login'.```

Crazybus:
None of your changes should be affecting the integration tests. The template testing passed though so good enough https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+template-testing/73/. I'll look into the flakiness of the other tests but that's unrelated to your changes. 

Thank you again for adding this in!

tetianakravchenko:
@Crazybus any updates on when those changes will be released?

PR #207, opened 2019-07-06
[elasticsearch] Add service account and configure helpers.tpl

naseemkullah:
Signed-off-by: Naseem <naseemkullah@gmail.com>

- [x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [x] README.md updated with any new values or changes
- [ ] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

Fixes #61

PR #208, opened 2019-07-06
[kibana] Make service account configurable, create one by default, modify help…

naseemkullah:
…ers.tpl


kibana/README.md
```diff
@@ -55,7 +55,8 @@ helm install --name kibana elastic/kibana --version 7.2.0 --set imageTag=7.2.0
...
-| `serviceAccount`          | Allows you to overwrite the "default" [serviceAccount](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) for the pod                                                                                                                                                                                                        | `[]`                                                                                                                      |
+| `serviceAccount.create`    | Create service account               | `true`                                                                                                                                   |
+| `serviceAccount.name`    | The name of the ServiceAccount to use. If not set and create is true, a name is generated using the fullname template               | ``                                                                                                                                   |
```
Crazybus: `serviceAccount` is already the name. Moving it to `serviceAccount.name` is going to be a breaking change. To do this in a backwards compatible way I propose using the same setup that filebeat and metricbeat are using with `managedServiceAccount` determining whether or not to manage the actual service account.

naseemkullah: Alright I've made some adjustments, but FWIW I think it's worth making this breaking change, because now we have `serviceAccount.name` and `managedServiceAccount.name`(with `managedServiceAccount.create` set to `false`) that achieve the exact same thing.
I would really prefer going with original PR and sticking with what's standard in so many other charts.

Crazybus: Updating values to match the current standard being used is not a good enough reason to make a breaking change. We can totally do it for the next major release but that is still a while off. Or worse, it will cause a subtle breakage for anyone upgrading and not reading the readme for every single release. I find it very important that users can continue to upgrade to the latest version of the chart to take advantage of new features and fixes without needing to review every single change.

naseemkullah: Fair enough!

kibana/values.yaml
```diff
@@ -65,7 +65,12 @@ securityContext:
   runAsNonRoot: true
   runAsUser: 1000
 
-serviceAccount: ""
+serviceAccount:
```
Crazybus:
```
==> Linting ./
[ERROR] templates/: render error in "kibana/templates/serviceaccount.yaml": template: kibana/templates/serviceaccount.yaml:1:14: executing "kibana/templates/serviceaccount.yaml" at <.Values.managedServiceAccount.create>: map has no entry for key "managedServiceAccount"
```

This value is still referring to the previous changes so still needs to be updated for `managedServiceAccount`.

kibana/templates/deployment.yaml
```diff
@@ -37,6 +37,8 @@ spec:
 {{ toYaml .Values.podSecurityContext | indent 8 }}
       {{- if .Values.serviceAccount }}
       serviceAccount: {{ .Values.serviceAccount }}
+      {{- else if .Values.managedServiceAccount }}
```
Crazybus: I think this one needs to be looking at the `.create` key instead.
```suggestion
      {{- else if .Values.managedServiceAccount.create }}
```

kibana/values.yaml
```diff
@@ -65,7 +65,12 @@ securityContext:
...
+  # Specifies whether a service account should be created
+  create: true
```
Crazybus: This should be defaulting to false to preserve backwards compatibility.

kibana/README.md
```diff
@@ -54,6 +54,9 @@ helm install --name kibana elastic/kibana --set imageTag=7.3.0
...
+| `managedServiceAccount.name`     | The name of the ServiceAccount to use. If not set and create is true, a name is generated using the fullname template                                                                                                                                                                                                                                   | ``                                                                                                                        |
+| `serviceAccount`          | Allows you to overwrite the "default" [serviceAccount](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) for the pod                                                                                                                                                                                                        | `[]`                                                                                                                      |
```
Crazybus: This is a duplicate line since it is already in the readme

jmlrt:
jenkins test this please

jmlrt:
Hi @naseemkullah,
Following https://github.com/elastic/helm-charts/pull/207#issuecomment-534282406, service account creation has already been implemented in Elasticsearch Chart by https://github.com/elastic/helm-charts/pull/265 and for consistency reasons, we expect it to be implemented in the same way in every Elastic Helm Charts.
Thank you for your great work on this PR.
I'll keep it open until we have a PR merged with an implementation similar to what's done in https://github.com/elastic/helm-charts/pull/265 for Elasticsearch. Don't hesitate reopen another PR for the new labels creation or if you want to provide https://github.com/elastic/helm-charts/pull/265 implementation for Kibana and the Beats Charts.

naseemkullah:
Hi @jmlrt just a small note with regards to #265, if I understand correctly the rbac key also creates a service account, according to helm documentation it is better to have these created under separate keys: https://helm.sh/docs/chart_best_practices/#yaml-configuration

jmlrt:
Hi @naseemkullah, sorry for the very late answer.

Thanks for your link.

Also following Helm Best Practices is important, our main priority with these Helm charts are   "ensuring that we don't introduce any breaking changes so upgrading from a version of the chart to a newer shouldn't break running Helm releases" (we'll have to add some one day, but we prefer not currently) and "ensuring that we provide consistency between all our charts (Elasticsearch, Kibana, Logstash, Filebeat and Metricbeat) so that a user shouldn't have different ways to do something in one chart or another".

As #265 was merged before for Elasticsearch, we prefer not change our way to manage RBAC and PSP unless we find some way to do it which won't break for existing users during `helm upgrade` and we prefer not introduce a different way to manage service accounts in other charts to avoid inconsistencies.

naseemkullah:
Very well @jmlrt, thanks for the explanation! Closing.

PR #210, opened 2019-07-07
Added kibana pod annotations

cliedeman:


kibana/tests/kibana_test.py
```diff
@@ -333,3 +333,11 @@ def test_override_the_serverHost():
...
     assert c['env'][1]['name'] == 'SERVER_HOST'
     assert c['env'][1]['value'] == 'localhost'
+
+  def test_adding_pod_annotations():
+    config = '''
+podAnnotations:
+  iam.amazonaws.com/role: es-role
+'''
+    r = helm_template(config)
+    assert r['deployment'][uname]['spec']['containers'][0]['annotations']['iam.amazonaws.com/role'] == 'es-role'
```
Crazybus: The key you are referring to here is for the annotations for the first container and not the entire pod. I think you want `assert r['deployment'][name]['spec']['template']['metadata']['annotations']['iam.amazonaws.com/role'] == 'es-role'`

kibana/tests/kibana_test.py
```diff
@@ -333,3 +333,11 @@ def test_override_the_serverHost():
     c = r['deployment'][name]['spec']['template']['spec']['containers'][0]
     assert c['env'][1]['name'] == 'SERVER_HOST'
     assert c['env'][1]['value'] == 'localhost'
+
+  def test_adding_pod_annotations():
```
Crazybus: There is some extra whitespace here which causes python to not run.

kibana/tests/kibana_test.py
cliedeman: That makes sense. Updated. I did this blind and was hoping jenkins would help me along 😄

kibana/tests/kibana_test.py
cliedeman: Fixed

Crazybus:
jenkins test this please

PR #220, opened 2019-07-10
Add option to disable sysctlInitContainer

JorisAndrade:
Following https://github.com/elastic/helm-charts/pull/132 that looked pretty abandoned. 
I just fixed the comments of this PR and added an entry to the README.md


elasticsearch/README.md
```diff
@@ -108,7 +108,8 @@ helm install --name elasticsearch elastic/elasticsearch --version 7.2.0 --set im
...
+| `lifecycle`                 | Allows you to add lifecycle configuration. See [values.yaml](./values.yaml) for an example of the formatting.                                                                                                                                                                                                             | `{}`                                                                                                                      |
+| `sysctlInitContainer.enabled` | Allows you to disable the sysctlInitContainer if you are not allowed to run privileged:true.                                                                                                                                                                                                                            | `false`                                                                                                                   |
```
Crazybus: The default value is `true` but you have put `false` here.

elasticsearch/tests/elasticsearch_test.py
```diff
@@ -82,6 +82,10 @@ def test_defaults():
             'name': 'node.ingest',
             'value': 'true'
         },
+        {
+            'name': 'sysctlInitContainer.enabled',
```
Crazybus: I'm not going to run CI because I already know that this test will fail. This test is checking to see if there is an environment variable called `sysctlInitContainer.enabled` that is set to `true`. A test similar to this is what you are looking for: https://github.com/elastic/helm-charts/blob/288e5f72e76fd46254c43d7880b49957dc169b45/elasticsearch/tests/elasticsearch_test.py#L326-L335

elasticsearch/README.md
Crazybus: > if you are not allowed to run privileged:true.

This is only one of the reasons why you might want to disable this. You might already have this set on the node level, or you might (when an option is added) be using the default built in sysctl integration that Kubernetes offers. I think this part should just be removed to avoid any confusion.

Crazybus: Could you also move it next to the `sysctlVmMaxMapCount` value in this readme and in values.yaml

elasticsearch/tests/elasticsearch_test.py
JorisAndrade: added 2 test :)

elasticsearch/README.md
JorisAndrade: done!

Crazybus:
jenkins test this please

JorisAndrade:
> LGTM! Thanks a lot for picking this up and adding in the tests ⭐️

My pleasure, won't hesitate to create other PRs if we are stuck on something :)

PR #237, opened 2019-07-29
publishNotReadyAddresses not working

MichaelSp:
according to https://github.com/kubernetes/kubernetes/issues/58662 the `publishNotReadyAddresses: true` is not working. Instead they suggest this annotation as workaround: `service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"`

### DIFF
--- a/elasticsearch/templates/service.yaml
+++ b/elasticsearch/templates/service.yaml
@@ -37,6 +37,8 @@ metadata:
     release: {{ .Release.Name | quote }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     app: "{{ template "uname" . }}"
+  annotations:
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
 spec:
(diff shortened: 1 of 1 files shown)


Crazybus:
Which version of Kubernetes are you running? When we deprecated this option I checked to make sure that it was supported by 1.8 https://github.com/elastic/helm-charts/pull/123#discussion_r281612632 and it was. However that issue you linked to makes it sound like this option was never properly implemented even though the changelog made it sound like it was. I'm curious to know which version you are running since I would hope that this fix was backported to the latest 1.8 release too. 

jenkins test this please

Crazybus:
I confirmed that this is indeed broken on the latest 1.8 cluster running in minikube. 1.8 isn't part of the automated testing suite anymore since GKE no longer has the version available. 

I'll make sure to manually test anything that could affect backwards compatibility for 1.8 from no on (instead of blindly trusting the changelogs). 

Thank you for finding and fixing this!

MichaelSp:
I'm running `v1.10.11`

Issue #235, opened 2019-07-26
Filebeat: wrong type for value, extraVolumes/extraVolumeMounts


spencergilbert:
**Chart version:**
7.2.1-0

**Kubernetes version:**
N/A

**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
Rancher

**Helm Version:**
2.9.1 and 3.0.0-alpha.2

**Describe the bug:**
When `helm template`ing the filebeat chart I'm getting a type error for values passed in under extraVolumes and extraVolumeMounts 

**Steps to reproduce:**

1. apply the volume config below!

**Expected behavior:**
Pass in extra volumes and mount them without error

**Provide logs and/or server output (if relevant):**
`filebeat/templates/daemonset.yaml:67:14: executing "filebeat/templates/daemonset.yaml" at <.Values.extraVolumes>: wrong type for value; expected string; got []interface {}`
**Any additional context:**

```
extraVolumeMounts:
  - name: varlog
    mounthPath: /var/log
    readOnly: true
  - name: rkelog
    mountPath: /var/lib/rancher/rke/log
    reandOnly: true
  # - name: extras
  #   mountPath: /usr/share/extras
  #   readOnly: true

extraVolumes:
  - name: varlog
    hostPath:
      path: /var/log
  - name: rkelog
    hostPath:
      path: /var/lib/rancher/rke/log
  # - name: extras
  #   emptyDir: {}
```

zeysh:
Hi,

Same problem for me. I have tried it for containerd logs.

```
extraVolumeMounts:
  - name: containerd
    mountPath: /var/log/containerd
    readOnly: true
```

```
==> Linting .
[ERROR] templates/: render error in "filebeat/templates/daemonset.yaml": template: filebeat/templates/daemonset.yaml:138:14: executing "filebeat/templates/daemonset.yaml" at <.Values.extraVolumeMounts>: wrong type for value; expected string; got []interface {}

Error: 1 chart(s) linted, 1 chart(s) failed
```

spencergilbert:
@zeysh i talked to a team mate of mine and he pointed out if you put a `|` in front the valuemount key it should work..

```
extraVolumeMounts: | 
  - name: containerd
    mountPath: /var/log/containerd
    readOnly: true
```

Crazybus:
[...]

PR #238, opened 2019-07-30
[elasticsearch] Update documentation and defaults for tmpl values

Crazybus:
Closes: #235


Crazybus:
> Do the other charts like Filebeat need this change as well? I noted that the original report was for Filebeat and Filebeat chart has default values of empty arrays as well [].

Absolutely yes. Thank you for catching this, I have updated the PR with the changes for metricbeat and filebeat (not implemented in Kibana). PTAL!

PR #240, opened 2019-07-31
[meta] Add dependency download to release script

Crazybus:
This is needed for metricbeat to download the child charts


Crazybus:
I aborted the CI build since this script is not run as part of CI.

PR #241, opened 2019-08-01
 Change antiaffinity match value to fixed "kibana"

oneoneonepig:
Fixing "antiAffinity" value is not used in deployment.
README.md accepts value soft/hard, while actual deployment.yaml file not using the value, using the whole YAML instead. 

The whole section is copied from elasticsearch's chart. Since no uname is defined, the antiaffinity match value is configured to fixed value "kibana"

### DIFF
--- a/kibana/templates/deployment.yaml
+++ b/kibana/templates/deployment.yaml
@@ -52,10 +52,39 @@ spec:
     {{- with .Values.nodeSelector }}
       nodeSelector:
 {{ toYaml . | indent 8 }}
-    {{- end }}
-    {{- with .Values.affinity }}
+      {{- end }}
+      {{- if or (eq .Values.antiAffinity "hard") (eq .Values.antiAffinity "soft") .Values.nodeAffinity }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
+      {{- end }}
       affinity:
-{{ toYaml . | indent 8 }}
+      {{- end }}
+      {{- if eq .Values.antiAffinity "hard" }}
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - "{{ template "uname" .}}"
+            topologyKey: {{ .Values.antiAffinityTopologyKey }}
+      {{- else if eq .Values.antiAffinity "soft" }}
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 1
+            podAffinityTerm:
+              topologyKey: {{ .Values.antiAffinityTopologyKey }}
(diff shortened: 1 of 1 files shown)


kibana/templates/deployment.yaml
```diff
@@ -52,10 +52,39 @@ spec:
...
+      {{- end }}
+      {{- if eq .Values.antiAffinity "hard" }}
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - kibana
```
Crazybus: This value is not unique per deployment. If someone has multiple Kibana deployments in the same namespace then this will fail (our integration testing does this so will also fail here).

kibana/templates/deployment.yaml
```diff
@@ -52,10 +52,39 @@ spec:
     {{- with .Values.nodeSelector }}
       nodeSelector:
 {{ toYaml . | indent 8 }}
-    {{- end }}
-    {{- with .Values.affinity }}
+      {{- end }}
+      {{- if or (eq .Values.antiAffinity "hard") (eq .Values.antiAffinity "soft") .Values.nodeAffinity }}
+      {{- if .Values.priorityClassName }}
```
Crazybus: This is already defined at https://github.com/elastic/helm-charts/blob/741de11d2226bc07957cbdcc11660169650e7099/kibana/templates/deployment.yaml#L33-L35. Did you mean to duplicate this?

kibana/templates/deployment.yaml
```diff
@@ -52,10 +52,39 @@ spec:
...
+            podAffinityTerm:
+              topologyKey: {{ .Values.antiAffinityTopologyKey }}
+              labelSelector:
+                matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                  - kibana
+      {{- end }}
+      {{- with .Values.nodeAffinity }}
```
Crazybus: Can you add a default value for this (strict linting will fail otherwise) and an entry to the readme and some basic template tests in `kibana/tests/kibana_test.py `

kibana/templates/deployment.yaml
```diff
@@ -52,10 +52,39 @@ spec:
     {{- with .Values.nodeSelector }}
       nodeSelector:
 {{ toYaml . | indent 8 }}
-    {{- end }}
-    {{- with .Values.affinity }}
```
Crazybus: The problem here is that the documentation is wrong, not the template itself. This behavior works and actually has template testing for it defined here: https://github.com/elastic/helm-charts/blob/007d6e4096c4419200b7c52c56781e9e1974c369/kibana/tests/kibana_test.py#L117-L133. Switching this over to match the Elasticsearch chart makes sense, however this would make it a breaking change for anybody that is correctly configuring this (by ignoring the wrong documentation). So I think the way forward here is to update the documentation not the template.

PR #245, opened 2019-08-02
Fixed indent on elasticsearch extraVolumes tpl. Was causing parsing errors.

skitle:
Fixed indent on elasticsearch extraVolumes tpl. Was causing parsing errors.

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -112,7 +112,7 @@ spec:
             name: {{ template "uname" . }}-config
         {{- end }}
       {{- if .Values.extraVolumes }}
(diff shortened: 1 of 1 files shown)


skitle:
Updated CLA

Crazybus:
Could you post what your configuration looks like for this? We have a test for this which passes with and without this indentation. In the YAML spec both should be and are valid so I would be interested to see what configuration you used to break this (as it might affect other values).

This is the test:

https://github.com/elastic/helm-charts/blob/24ac48ac81c43c137e482c47f3a1bf616f49e40d/elasticsearch/tests/elasticsearch_test.py#L309-L323

Crazybus:
jenkins test this please

PR #109, opened 2019-04-26
add support for k8s PriorityClass

lancespeelmon:

PR #248, opened 2019-08-06
Add priorityClassName to filebeat chart

tanordheim:
This adds the support for `priorityClassName` on the filebeat chart, the same way that got added to the elasticsearch and kibana charts in #109. As in the other PR, it defaults to no `priorityClassName` to avoid breaking anything.


filebeat/README.md
```diff
@@ -66,6 +66,7 @@ helm install --name filebeat elastic/filebeat --set imageTag=7.3.0
...
 | `affinity`               | Configurable [affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity)                                                                                                                                                      | `{}`                                                                                                                      |
+| `priorityClassName`      | The [name of the PriorityClass](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass). No default is supplied as the PriorityClass must be created first.                                                                               | ``                                                                                                                        |
```
Crazybus: This should be an empty string not an empty value (doesn't really matter though).

```suggestion
| `priorityClassName`      | The [name of the PriorityClass](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass). No default is supplied as the PriorityClass must be created first.                                                                               | `""`                                                                                                                        |
```

Crazybus:
jenkins test this please

filebeat/README.md
tanordheim: I just went with how this was configured in the other charts that also have `priorityClassName` available; I'm happy to change it if you wish?

Crazybus: Go for it! If you could fix it in the other charts too that would be great :)

tanordheim: I'll handle it in a separate PR.

PR #250, opened 2019-08-07
Update priorityClassName default values in READMEs

tanordheim:
Updates READMEs for the elasticsearch, filebeat and kibana charts to clarify that the default value of `priorityClassName` is an empty string, re a discussion in #248.

I've made the change in one commit per chart, let me know if you'd rather want these squashed into a single commit.


tanordheim:
> Thank you for so quickly following up...it was almost too fast so I think you might be a robot ;)

Issue #15, opened 2018-12-19
Openshift support


gsaslis:
Hello there, 

I was looking for the best way to deploy elasticsearch / kibana on openshift and it looks like I could use some help.  : )

I tried adapting these charts and [kept some notes with the problems that came up](https://gist.github.com/gsaslis/b12c57a9b34ced5bff0d1ed1c3e20d35).

I would welcome suggestions (even for a completely different method of deploying into openshift - e.g. openshift templates / operator) and am happy to help with any questions you might have so we can figure something out (I have an openshift cluster standing by).

Thanks!

Crazybus:
Thank you so much for taking the time to test this out and writing up detailed notes. Adding support for other platforms (like Openshift) is planned for the future but we are currently only testing and supporting Google Kubernetes Engine (GKE).

Just a heads up, I have never used Openshift before so this will be my first time trying to figure this stuff out too! I don't currently have access to an Openshift cluster to help debug but in the future I hope to come back to this to get it properly working if you can't figure it out with my comments. 

> I couldn’t use Helm itself due to permissions issues related to: https://github.com/helm/helm/issues/1918

Is the problem that you don't want to use tiller because of the permissions required? Or is it that you are unable to get it working on Openshift? A quick google makes it sound like helm should work just fine on openshift as long as you install it per namespace. They even have an [official blog post](https://blog.openshift.com/getting-started-helm-openshift/) about how to install it. 


> the initContainer tries to run as root.

There is a [beta feature](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/) in Kubernetes 1.12 to set sysctls natively through Kubernetes. It looks like `vm.*` would be considered "unsafe" and need to be allowed on a cluster level. [Openshift](https://docs.openshift.com/container-platform/3.3/admin_guide/sysctls.html#setting-sysctls-for-a-pod) also seems to support it though it still looks to be the alpha version. I can certainly see having the root init container being optional so that users can disable it and configure the needed sysctl values via annotations instead since I can imagine other users wanting to use this new feature too.

> the fsGroup seems to have an invalid value. I am not sure what that should be. Deleting the fsGroup + securityContext does allow statefulset to be created, BUT, again, that’s probably there for a reason (haven’t looked into that in detail yet)

[...]

PR #263, opened 2019-08-22
Add working examples for running Elasticsearch and Kibana on ope…

Crazybus:
Closes: #15

We don't currently have automated testing for openshift clusters however
these have been manually tested and confirmed to work. For now having
an example on how to do this is nice but in the future we can consider
adding a value to pragmatically disable these features with something
like `openshift: true`

Special thanks to @geogdog for getting these working and providing the
correct changes needed.

### DIFF
--- a/elasticsearch/examples/openshift/Makefile
+++ b/elasticsearch/examples/openshift/Makefile
@@ -1,0 +1,15 @@
+default: test
+include ../../../helpers/examples.mk
+
+RELEASE := elasticsearch
+
+template:
+	helm template --values ./values.yaml ../../
+
+install:
+	helm upgrade --wait --timeout=600 --install $(RELEASE) --values ./values.yaml ../../ ; \
+
+test: install goss
+
+purge:
+	helm del --purge $(RELEASE)
--- a/elasticsearch/examples/openshift/Makefile
+++ b/elasticsearch/examples/openshift/Makefile
@@ -7,7 +7,7 @@ template:
 	helm template --values ./values.yaml ../../
 
 install:
-	helm upgrade --wait --timeout=600 --install $(RELEASE) --values ./values.yaml ../../ ; \
+	helm upgrade --wait --timeout=600 --install $(RELEASE) --values ./values.yaml ../../
 
 test: install goss
 
--- a/kibana/examples/openshift/Makefile
+++ b/kibana/examples/openshift/Makefile
@@ -7,7 +7,7 @@ template:
 	helm template --values ./values.yml ../../
 
 install: 
-	helm upgrade --wait --timeout=600 --install --values ./values.yml $(RELEASE) ../../ ; \
+	helm upgrade --wait --timeout=600 --install --values ./values.yml $(RELEASE) ../../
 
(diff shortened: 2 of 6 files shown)


kibana/examples/openshift/values.yml
```diff
@@ -0,0 +1,4 @@
+---
+
+fsGroup: null
+runAsUser: null
```
jmlrt: Shouldn't we have the same structure as in `elasticsearch/examples/openshift/values.yaml`?
```
securityContext:
  runAsUser: null

podSecurityContext:
  fsGroup: null
```

elasticsearch/examples/openshift/Makefile
```diff
@@ -0,0 +1,15 @@
+default: test
+include ../../../helpers/examples.mk
+
+RELEASE := elasticsearch
+
+template:
+	helm template --values ./values.yaml ../../
+
+install:
+	helm upgrade --wait --timeout=600 --install $(RELEASE) --values ./values.yaml ../../ ; \
```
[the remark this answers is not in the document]
Crazybus: fixed! Guessing it was leftover from a copy paste where it is doing multiple commands.

kibana/examples/openshift/values.yml
Crazybus: It would be nice indeed, but trying to keep the charts identical for every change is quite the challenge. For backwards compatibility reasons the Elasticsearch chart does actually still support `fsGroup` directly as per https://github.com/elastic/helm-charts/pull/171. These changes could also be backported to Kibana. Wherever possible I have been trying to keep changes in sync and to follow the same style as other charts but this one slipped in.

jmlrt: Isn't this already the case in [kibana/values.yaml#L57-L66](https://github.com/elastic/helm-charts/blob/master/kibana/values.yaml#L57-L66)?

```
podSecurityContext:
  fsGroup: 1000

securityContext:
  capabilities:
    drop:
    - ALL
  # readOnlyRootFilesystem: true
  runAsNonRoot: true
  runAsUser: 1000
```

Crazybus: You are 100% right. I could have sworn this was working before my holiday. When I ran `make template` just now it is clearly not doing anything. Just pushed a new commit using the correct values.

kibana/examples/openshift/Makefile
```diff
@@ -0,0 +1,15 @@
+default: test
+include ../../../helpers/examples.mk
+
+RELEASE := kibana
+
+template:
+	helm template --values ./values.yml ../../
+
+install: 
+	helm upgrade --wait --timeout=600 --install --values ./values.yml $(RELEASE) ../../ ; \
```
jmlrt:
```suggestion
	helm upgrade --wait --timeout=600 --install --values ./values.yml $(RELEASE) ../../
```

Crazybus: Fixed! Weird how committing your suggestion still requires you to re-review it haha

PR #270, opened 2019-09-03
ES Variable Port Name

GreenKnight15:
Template the port name values and set a default value

This is requested because running ES in an Istio Service Mesh with security enabled the port name needs to be different so Istio handles the traffic differently. Istio can not handle https traffic when it is expecting http traffic via the port name. https://istio.io/docs/setup/kubernetes/additional-setup/requirements/ 

- [X] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`


Crazybus:
jenkins test this please

Crazybus:
jenkins test this please

GreenKnight15:
@Crazybus Thanks for reviewing, I made some changes to get the build to pass. Also when I run `helm lint --strict` I run into other errors unrelated to my changes. 

> elasticsearch/templates/service.yaml" at <.Values.service.nodePort>: map has no entry for key "nodePort"

Also to answer your question and to expand on the PR description, we are using the all of the ELK helm charts (which are a work of art imo) behind an Isito service mesh which enables mTLS traffic between pods. 

After enabling the elastic security features including SSL istio could not interpret the traffic and the masters could never cluster.  I needed to changes how istio handles the traffic between pods which is taken care of by updating the service port name as described [here](https://istio.io/docs/setup/additional-setup/requirements/). 

Following [this issue](https://github.com/istio/istio/issues/10573) I updated the service port name to be `tcp-http` which changes how istio handles the traffic which fixed the issue, and because elastic has ssl enable traffic is still secure.

GreenKnight15:
jenkins test this please

Crazybus:
> we are using the all of the ELK helm charts (which are a work of art imo)

Thank you so much, it really means a lot :)

> After enabling the elastic security features including SSL istio could not interpret the traffic and the masters could never cluster. I needed to changes how istio handles the traffic between pods which is taken care of by updating the service port name as described here.

> Following this issue I updated the service port name to be tcp-http which changes how istio handles the traffic which fixed the issue, and because elastic has ssl enable traffic is still secure.

Perfect! Thank you so much for explaining this. In the next major release we could consider using istio compatible naming for these ports. However I'm sure every service mesh has their own standards so they will always need to be configurable.

elasticsearch/templates/service.yaml
```diff
@@ -18,13 +18,13 @@ spec:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     app: "{{ template "uname" . }}"
   ports:
-  - name: http
+  - name: {{ .Values.service.httpPortName | default "http" }}
```
Crazybus: The defaults can be removed from here now that you have added the defaults into the values.yaml.

```suggestion
  - name: {{ .Values.service.httpPortName }}
```

Crazybus:
jenkins test this please

GreenKnight15:
Thanks again, I look forward to contributing again in the future.

PR #274, opened 2019-09-10
Adding Example for Kubernetes KIND

salaboy:
- [x ] README.md updated with any new values or changes

This PR adds support for running elastic for development purposes on Kubernetes KIND (https://github.com/kubernetes-sigs/kind) which for some reason require the directory to be created before using it for elasticsearch data. Hence the use of an initContainer.

### DIFF
--- a/elasticsearch/examples/kubernetes-kind/values.yaml
+++ b/elasticsearch/examples/kubernetes-kind/values.yaml
@@ -1,0 +1,37 @@
+---
+# Permit co-located instances for solitary minikube virtual machines.
+antiAffinity: "soft"
+
+# Shrink default JVM heap.
+esJavaOpts: "-Xmx128m -Xms128m"
+
+# Allocate smaller chunks of memory per pod.
+resources:
+  requests:
+    cpu: "100m"
+    memory: "512M"
+  limits:
+    cpu: "1000m"
+    memory: "512M"
+
+# Request smaller persistent volumes.
+volumeClaimTemplate:
+  accessModes: [ "ReadWriteOnce" ]
+  storageClassName: "hostpath"
+  resources:
+    requests:
+      storage: 100M
+extraInitContainers: |
+  - name: create
+    image: busybox:1.28
+    command: ['mkdir', '/usr/share/elasticsearch/data/nodes/']  
+    volumeMounts:
+    - mountPath: /usr/share/elasticsearch/data
+      name: elasticsearch-master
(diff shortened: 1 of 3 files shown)


elasticsearch/examples/kubernetes-kind/values.yaml
```diff
@@ -0,0 +1,37 @@
...
+    memory: "512M"
+
+# Request smaller persistent volumes.
+volumeClaimTemplate:
+  accessModes: [ "ReadWriteOnce" ]
+  storageClassName: "hostpath"
+  resources:
+    requests:
+      storage: 100M
+extraInitContainers: |
```
Crazybus: Could you explain why these extra init containers are needed for a kind deployment? The chart is already correctly setting the permissions and the elasticsearch process should be creating those directories.

salaboy: @Crazybus yeah.. that is needed for KIND.. for some reason the container doesn't create that by default and the whole deployment was failing. You can give it a try and see if you find the reason why the container is failing to do so.. I couldn't see anything in the logs except for an exception saying that it didn't had access to write the files in the `/usr/share/elasticsearch/data/nodes/` directory. Once the directory is created, everything works as expected.

Crazybus: Thanks, I'll certainly give it a go to see if I can reproduce it. Seems like a bit of a shame that kind doesn't seem to be compatible with the way that normal kubernetes works though :(. It makes sense that resources and node anti affinity need to be relaxed but this makes it impossible to properly test with kind.

salaboy: It should be super straight forward to reproduce it and find the cause if you can debug the ElasticSearch container. But for now that is needed.

salaboy: @Crazybus also I notice that for KIND this is required:  `storageClassName: "standard"`

Crazybus: Sorry that I haven't gotten around to testing this out just yet. Did you also try just leaving `storageClassName` out completely? That should just use whatever the default is set to which is probably "standard"

salaboy: that is true.. I am removing that line completely

jmlrt:
jenkins test this please

Issue #283, opened 2019-09-16
Kube-State-Metrics are not processed (MetricBeat)


julianrubisch:
**Chart version:** 7.3.0

**Kubernetes version:** 1.12.8

**Kubernetes provider:** Azure

**Helm Version:**
```
Client: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
```

**`helm get release` output**

```
REVISION: 1
RELEASED: Mon Sep 16 15:09:44 2019
CHART: metricbeat-7.3.0
USER-SUPPLIED VALUES:
{}

COMPUTED VALUES:
extraEnvs: []
extraVolumeMounts: []
extraVolumes: []
fullnameOverride: ""
hostPathRoot: /var/lib
image: docker.elastic.co/beats/metricbeat
imagePullPolicy: IfNotPresent
imagePullSecrets: []
imageTag: 7.3.0
kube-state-metrics:
  affinity: {}
  collectors:
    certificatesigningrequests: true
    configmaps: true
    cronjobs: true
    daemonsets: true
    deployments: true
    endpoints: true
    horizontalpodautoscalers: true
    ingresses: true
    jobs: true
    limitranges: true
    namespaces: true
    nodes: true
    persistentvolumeclaims: true
    persistentvolumes: true
    poddisruptionbudgets: true
    pods: true
    replicasets: true
    replicationcontrollers: true
    resourcequotas: true
    secrets: true
    services: true
    statefulsets: true
  global: {}
  hostNetwork: false
  image:
    pullPolicy: IfNotPresent
    repository: quay.io/coreos/kube-state-metrics
    tag: v1.6.0
  nodeSelector: {}
  podAnnotations: {}
  podSecurityPolicy:
    annotations: {}
    enabled: false
  prometheus:
    monitor:
      additionalLabels: {}
      enabled: false
      namespace: ""
  prometheusScrape: true
  rbac:
    create: true
  replicas: 1
  securityContext:
    enabled: true
    fsGroup: 65534
    runAsUser: 65534
  service:
    loadBalancerIP: ""
    nodePort: 0
    port: 8080
    type: ClusterIP
  serviceAccount:
    create: true
    imagePullSecrets: []
    name: null
  tolerations: []
[...]

PR #289, opened 2019-09-18
[metricbeat] Remove default kube static metrics host to avoid co…

Crazybus:
This default value is never used since the environment variable is
always being set for the deployment. See
https://github.com/elastic/helm-charts/issues/283#issuecomment-532708548
for more information.

### DIFF
--- a/metricbeat/examples/security/values.yaml
+++ b/metricbeat/examples/security/values.yaml
@@ -61,7 +61,7 @@ metricbeatConfig:
         - state_pod
         - state_container
       period: 10s
-      hosts: ["${KUBE_STATE_METRICS_HOSTS:kube-state-metrics:8080}"]
+      hosts: ["${KUBE_STATE_METRICS_HOSTS}"]
     output.elasticsearch:
       username: '${ELASTICSEARCH_USERNAME}'
       password: '${ELASTICSEARCH_PASSWORD}'
--- a/metricbeat/values.yaml
+++ b/metricbeat/values.yaml
@@ -58,7 +58,7 @@ metricbeatConfig:
         - state_pod
         - state_container
       period: 10s
-      hosts: ["${KUBE_STATE_METRICS_HOSTS:kube-state-metrics:8080}"]
+      hosts: ["${KUBE_STATE_METRICS_HOSTS}"]
     output.elasticsearch:
       hosts: '${ELASTICSEARCH_HOSTS:elasticsearch-master:9200}'
 


Crazybus:
Kube state metrics is installed as a child chart (https://github.com/elastic/helm-charts/blob/master/metricbeat/requirements.yaml) which is used to scrape the Kubernetes API metrics for metricsbeat. Does that explain why you can't see where the kube state metrics service is defined? Or did you mean something else?

jmlrt:
> Kube state metrics is installed as a child chart (https://github.com/elastic/helm-charts/blob/master/metricbeat/requirements.yaml)

Thanks I didn't check `requirements.yaml`.

> Or did you mean something else?

Well as I understand the line `value: "$({{ .Release.Name | replace "-" "_" | upper }}_KUBE_STATE_METRICS_SERVICE_HOST):$({{ .Release.Name | replace "-" "_" | upper }}_KUBE_STATE_METRICS_SERVICE_PORT_HTTP)"`, Assuming that the release name is `metricbeat`, `KUBE_STATE_METRICS_HOSTS` environment variable is set to a concatenation of the value of `$METRICBEAT_KUBE_STATE_METRICS_SERVICE_HOST` environment variable and the value of `$METRICBEAT_KUBE_STATE_METRICS_SERVICE_PORT_HTTP` environment variable (with `:`between them).

However, I don't find in this Charts or in [kube-state-metric](https://github.com/helm/charts/blob/master/stable/kube-state-metrics) Charts where `$METRICBEAT_KUBE_STATE_METRICS_SERVICE_HOST` and `$METRICBEAT_KUBE_STATE_METRICS_SERVICE_PORT_HTTP` environment variables are defined.

Did I miss or misunderstood something?

Crazybus:
Ahh I get ya now. Kubernetes automatically creates these environment variables for all pods running in the namespace pointing to the service. So if there is a service called metricbeat-kube-state-metrics there will be an environment variable inside all pods in that namespace called `METRICBEAT_KUBE_STATE_METRICS_SERVICE_HOST` and `METRICBEAT_KUBE_STATE_METRICS_SERVICE_PORT_HTTP`

The reason that we use the environment variable instead of the service name in the config is that the service name is hardcoded in the default values. I think we could add go templating logic in values.yaml but this felt like the cleaner option because it means that all users can just refer to `KUBE_STATE_METRICS_HOSTS` in their metricbeat config without needing to copy paste the templating logic.

jmlrt:
OK now I understand, I forgot that Kubernetes was automatically setting these environment variable.
Thanks for the reminder.

iahmad-khan:
Hi,

Is there a way to use metrics-server instead of kube-state-metrics.

thanks

Crazybus:
Not that I know of. As far as I can tell Metricbeat only supports collecting the state_* metricsets from kube-state-metrics:

https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-module-kubernetes.html

https://www.elastic.co/guide/en/beats/metricbeat/current/running-on-kubernetes.html

If you would like this to be possible it would need to be a feature requests in https://github.com/elastic/beats. Since this repository is just the Helm chart installing Metricbeat.

iahmad-khan:
@Crazybus Thanks. I am asking this becuase kubernetes installations are coming with metrics-server by default ( at least the one i installed using RKE ). So it will make sense if they support the one with default instead of adding another one just for metricbeat.

Issue #284, opened 2019-09-16
[ElasticSearch] Service becomes unavailable on chart upgrade


haizz:
**Chart version:**
7.3.0

**Describe the bug:**
Chart has Service selector spec bound to specific chart version:
https://github.com/elastic/helm-charts/blob/a49995539e4347e8729437cfbb54299023c81306/elasticsearch/templates/service.yaml#L18

Which means that upgrading the chart removes older pods from the service and makes the service totally unavailable.

Crazybus:
That is a really great find! Especially since this comes from the default `helm create` template. Any idea if it is still present in the default template? If so I would expect that this would be affecting just about every helm chart created with it.

Crazybus:
This has been fixed in #290. Thank you again for finding it!

JoaquinScript:
This is still happening to me. I'm not sure if it is the same issue but it's the same symptom: after upgrade the pod (only one replica) never turns green and the last log its from it turning yellow from red. But if I delete the chart (and the pvc) and install again, everything works as expected.

Is there a chance I'm not using an updated chart? how can I make sure of that?

fatmcgav:
@JoaquinScript Are you able to open a new issue and provide the details requested?

We're specifically interested in the output from `kubectl get events` and the logs from the elasticsearch pods during the upgrade process.

Cheers

PR #290, opened 2019-09-18
[elasticsearch] Drop version from chart label in service

Crazybus:
As identified in #284 having the version in the label would cause the
service to be recreated during a chart version upgrade.

### DIFF
--- a/elasticsearch/templates/configmap.yaml
+++ b/elasticsearch/templates/configmap.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     heritage: {{ .Release.Service | quote }}
     release: {{ .Release.Name | quote }}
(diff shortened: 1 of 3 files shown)


tetianakravchenko:
Does it mean that to migrate to next chart version (7.4) it will be required to reinstall chart?

Crazybus:
> Does it mean that to migrate to the next chart version (7.4) it will be required to reinstall the chart?

@tetianakravchenko Certainly not, and we will never purposefully bring in any breaking changes requiring a reinstall. We have an "upgrade" test as part of our automated testing pipeline https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/upgrade/Makefile to ensure that each releases remains backwards compatible with the initial 7.0.0 release.

PR #272, opened 2019-09-04
use same env variable as application

rccrdpccl:
Signed-off-by: Riccardo Piccoli <riccardo.piccoli@softonic.com>

- [x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [ ] README.md updated with any new values or changes
- [ ] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

It appears that env vars used by kibana are `ELASTICSEARCH_USERNAME` and `ELASTICSEARCH_PASSWORD`, but readiness uses `ELASTIC_USERNAME` and `ELASTIC_PASSWORD`. This causes kibana to be unready, unless we also configure `ELASTIC_` vars.

PR #291, opened 2019-09-18
[kibana] Explicitly test for a 200 for readinessProbe

Crazybus:
If authentication wasn't configured correctly the Kibana readinessProbe
would still pass since `curl --fail` doesn't consider a redirect to be a
failure. If authentication fails Kibana would redirect you back to the
login page so this readinessProbe always passed for security clusters.
The tests are going to fail until #272 is merged and this PR has been
rebased on top of it.


Crazybus:
This one fails as expected now!

image

Crazybus:
Rebased on master now that #272 has been merged in. Lets see if the tests now pass properly.

PR #294, opened 2019-09-19
add in /var/log and /var/data for kube logs

ceastman-ibm:

### DIFF
--- a/filebeat/templates/daemonset.yaml
+++ b/filebeat/templates/daemonset.yaml
@@ -65,9 +65,15 @@ spec:
         hostPath:
           path: {{ .Values.hostPathRoot }}/{{ template "fullname" . }}-{{ .Release.Namespace }}-data
           type: DirectoryOrCreate
+      - name: vardata
+        hostPath:
+          path: /var/data/          
       - name: varlibdockercontainers
         hostPath:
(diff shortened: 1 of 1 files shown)


Crazybus:
`/var/log` makes sense to me and is also in the official filebeat examples https://github.com/elastic/beats/blob/3e0cf6c63c9b83e4efe55f5e08e8daaa53f1cfe6/deploy/kubernetes/filebeat/filebeat-daemonset.yaml#L62-L64

Can you explain why /var/data is also being mounted? 

Also: the status check is currently failing as you have not signed our [CLA](https://www.elastic.co/contributor-agreement)

ceastman-ibm:
there is a soft link in the /var/log directory that maps to /var/data that needs to be mounted also.

[root@filebeat-59sd4 log]# pwd
/var/log
[root@filebeat-59sd4 log]# ls -al pods
lrwxrwxrwx 1 root root 21 Sep 18 13:02 pods -> /var/data/kubeletlogs

@Crazybus How do I check if my company has an existing contributor agreement?

Crazybus:
> @Crazybus How do I check if my company has an existing contributor agreement?

As far as I know there aren't any company level CLAs. You will need to sign it with your own email company email address.

PR #297, opened 2019-09-20
[helm-tester] migrate to python 3

jmlrt:
Does what it says on the tin


jmlrt:
I don't know if we should stay with `-onbuild` Python Docker images as they seem to be deprecated (see https://github.com/docker-library/official-images/issues/2076).

But first test with `python:3.7` image is failing. I need to dig deeper at differences between `onbuild` and standard Docker images.

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

PR #298, opened 2019-09-20
Make it possible to override the endpoint template.

floretan:

We would like to use the elasticsearch chart as a subchart, where the name of the resource needs to match the release name (the common structure used by many charts, like the official mariadb chart). 

This was quite easy to do by overriding the helm template for `uname` and `masterService` (replacing their content with `{{ .Release.Name }}-elasticsearch`), but the `endpoints` template also needs to be overridden, and due to the way parameters are passed the trick above doesn't work.

This change simply passes moves the specification of `.Values` to the helm template, enabling overrides to access top-level parameters like `.Release.Name`. The behavior stays unchanged. This template is also not used anywhere else.

### DIFF
--- a/elasticsearch/templates/_helpers.tpl
+++ b/elasticsearch/templates/_helpers.tpl
@@ -28,8 +28,8 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 
 {{- define "endpoints" -}}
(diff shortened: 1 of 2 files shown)


Crazybus:
I have no idea why I did this, it doesn't make any sense :P Thank you for finding and fixing the issue!

jenkins test this please

PR #299, opened 2019-09-24
Use newer version for TF's google

flaper87:
The 2.14 version includes a fix for the issue linked below, which I was
hitting during my tests.

https://github.com/terraform-providers/terraform-provider-google/issues/4305


flaper87:
just noticed it was updated 5 days ago... too slow :P

Issue #300, opened 2019-09-24
metricbeat - kubernetes event.module gets "connection refused" accessing /stats/summary endpoint on EKS


robertgates55:
**Chart version:**
metricbeat 7.3.2

**Kubernetes version:**
1.13

**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
AWS EKS

**Helm Version:**
v2.14.3 (client & server)

**`helm get release` output**
```
REVISION: 3
RELEASED: Tue Sep 24 09:43:08 2019
CHART: metricbeat-7.3.2
USER-SUPPLIED VALUES:
extraEnvs:
- name: ELASTICSEARCH_USERNAME
  valueFrom:
    secretKeyRef:
      key: username
      name: elastic-credentials
- name: ELASTICSEARCH_PASSWORD
  valueFrom:
    secretKeyRef:
      key: password
      name: elastic-credentials
metricbeatConfig:
  kube-state-metrics-metricbeat.yml: |
    metricbeat.modules:
    - module: kubernetes
      enabled: true
      metricsets:
        - state_node
        - state_deployment
        - state_replicaset
        - state_pod
        - state_container
      period: 10s
      hosts: ["${KUBE_STATE_METRICS_HOSTS}"]
    output.elasticsearch:
      username: '${ELASTICSEARCH_USERNAME}'
      password: '${ELASTICSEARCH_PASSWORD}'
      protocol: https
      hosts: ["elasticsearch.qa.duco.services:443"]
  metricbeat.yml: |
    system:
      hostfs: /hostfs
    metricbeat.modules:
    - module: kubernetes
      metricsets:
        - container
        - node
        - pod
        - system
        - volume
      period: 30s
      host: "${NODE_NAME}"
      hosts: ["http://${HOSTNAME}:10250"]
      processors:
      - add_kubernetes_metadata:
          in_cluster: true
    - module: kubernetes
      enabled: true
      metricsets:
        - event
    - module: system
      period: 10s
      metricsets:
        - cpu
        - load
        - memory
        - network
        - process
        - process_summary
      processes: ['.*']
      process.include_top_n:
        by_cpu: 5
        by_memory: 5
    - module: system
      period: 1m
      metricsets:
        - filesystem
        - fsstat
      processors:
      - drop_event.when.regexp:
          system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
    output.elasticsearch:
[...]

PR #310, opened 2019-10-02
[metricbeat] Make cluster role rules configurable

Crazybus:
Fixes: #300


pbecotte:
Was just about to do this-

PR #314, opened 2019-10-05
Had to add a couple extra mounts to get MetricBeat to pick up all the…

pbecotte:
… metrics from the host nodes on Digital Ocean. In the process, noticed that the extraMount setting is configured different then the rest- changed it to be consistent. Not being a metricbeat expert, it is possible there was a better way to do this?


ostmike:
spent a few hours trying to work out why my extra mounts weren't working! Thanks for this. Would be great to see it merged.

jmlrt:
jenkins test this please

PR #320, opened 2019-10-11
optional labels for kibana ingress

paltryeffort:
I am running multiple IngressControllers and I need to add labels to my Ingress objects to select which controller to use.

Config example for values.yml:
```
ingress:
  enabled: true
  annotations:
    ingress.kubernetes.io/protocol: https
  labels:
    my-ingress: internal
```

### DIFF
--- a/kibana/templates/ingress.yaml
+++ b/kibana/templates/ingress.yaml
@@ -10,6 +10,9 @@ metadata:
     app: {{ .Chart.Name }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
+{{- with .Values.ingress.labels }}
+{{ toYaml . | indent 4 }}
+{{- end }}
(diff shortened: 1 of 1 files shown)


elasticcla:
Hi @paltryeffort, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/helm-charts/pull/320.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

pbecotte:
I think you need to add a default for that to the values.yaml file as well?

jmlrt:
Hi @paltryeffort,
Thanks for your PR.
Can you sign the CLA: https://www.elastic.co/contributor-agreement

jmlrt:
jenkins test this please

PR #321, opened 2019-10-12
use host networking for filebeat so that the stats have the correct n…

pbecotte:
…ode information

matches changes I made to metricbeat to get hostnames reporting properly on Digital Ocean


jmlrt:
jenkins test this please

jmlrt:
Hi @pbecotte, 
We got an issue with this PR and Filebeat 7.5.0. The fact is that enabling `hostNetwork` by default fail our tests in 7.5.0 (and even if tests are OK with 7.4.2, we are not in a good state as filebeat reported an error of pod already used but didn't fail).

We also have use cases where `hostNetwork` is not possible (OpenShift for example).

FYI we'll have to disable `hostNetwork` by default and allow enabling it in by setting `hostNetworking=true`.

pbecotte:
As I mentioned, I am not an elastic stack expert, so if that's necessary that's great :) It is important to mention though, without these changes nothing actually worked on digital ocean. Everything came up, but the metrics were all using container names as hostnames and none of the server level metrics were aggregating.

Would just ask to check that the metrics all look right with the changed config before making any decisions. The code that the comments implied were supposed to fix that didn't seem to work, and I iterated on it for a while before resorting to this approach.

jmlrt:
FYI it `HostNetwork` was disabled by default in https://github.com/elastic/helm-charts/commit/538af31822d3bbbac93b1a270493b4818a983666. In your case where you need it, you'll just need to set `HostNetworking: true` in your values.

Issue #267, opened 2019-08-30
Support fullnameOverride (Kibana chart)


ppawel:
**Describe the feature:**

Currently "fullname" helper macro provides name to service, ingress etc. and it uses release name, chart name (and optionally you can input "nameOverride") and this results in a "x-y" construct.

Allowing the user to specify `fullnameOverride` in values file will give additional flexibility with naming resources.

Examples:

https://github.com/helm/charts/blob/master/stable/grafana/templates/_helpers.tpl
https://github.com/helm/charts/blob/master/stable/prometheus/templates/_helpers.tpl

**Describe a specific use case for the feature:**

I want my service, ingress and other resources to be named just "kibana", without any prefix, like "monitoring-kibana" (because they are anyway in the "monitoring" namespace in my case so it's redundant).

davispw:
This is inconsistent with elasticsearch's helm chart (which does not use the Release name at all--the service is always just named "elasticsearch-master").

morganchristiansson:
I agree. My pod names are:

- kibana-kibana-xxx
- filebeat-filebeat-xxx
- metricbeat-metricbeat-xxx

Please fix-fix

jmlrt:
I'm closing this issue as it was implemented in #330

ppawel:
Thanks @tusciucalecs, you are a hero-hero!

PR #328, opened 2019-10-15
[Issue #267] Support fullnameOverride in kibana, filebeat, metricbeat…

tusciucalecs:
… charts


elasticcla:
Hi @tusciucalecs, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/helm-charts/pull/328.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Issue #69, opened 2019-02-28
Official Logstash Helm chart feature request


jjtdepaul:
Are you working on Logstash Helm chart by chance?  I found this one, but it's not working right:  https://github.com/helm/charts/tree/master/stable/logstash - it's got some issues....

jjtdepaul:
Perhaps you know of another stable HChart we could try?

Crazybus:
> Are you working on Logstash Helm chart by chance?

Not just yet, however it is planned. The next charts will be filebeat and metricbeat and then the plan is to start working on a logstash chart. 

> it's got some issues....
> Perhaps you know of another stable HChart we could try?

For now this appears to be the main chart so I would recommend opening an issue in the helm/charts repo in the meantime. 

I'll leave this issue open and update it once work starts on logstash.

james-crowley:
@jmlrt Any update on Logstash? Seems like Filebeat and Metricbeat both have been added to the repo.

james-crowley:
@Crazybus Any progress/insights you could give on the Logstash Chart status? To my knowledge it seems Filebeat and Metricbeat look like they had a code pushed for 7.2.1 release and thus they are added to the charts here. 

Is logstash the next chart slotted for development or has the plan changed?

Crazybus:
Correct, Metricbeat and Filebeat are done so Logstash is up next. When development begins we will be sure to leave a note in this issue with a link to the WIP pull request just like with the other charts.

jmlrt:
No I'm sorry, we don't have any ETA neither for development start nor for beta release, Logstash chart development is in our internal backlog and we'll start it as soon as we'll find time for it.

ceastman-ibm:
@jmlrt +1 on the updated logstash helm chart.

demisx:
@ceastman-ibm Please make sure to upvote the original issue from @jjtdepaul  instead.

jmlrt:
Hi all!
This is just to let you know that we will start Logstash Helm Chart development pretty soon.

ceastman-ibm:
@jmlrt if you need help testing it :) im ready

jmlrt:
Hi all!
FYI first draft PR for Logstash Helm Chart is #333

jmlrt:
[...]

Issue #61, opened 2019-02-14
[elasticsearch] Create or allow setting a serviceAccount to be used 


rwittrick:
Hopefully most clusters have something along the lines of [restricted](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies) for the default service account in a namespace.  By default, with no serviceAccount specified, the `default` serviceAccount is used.

Two values are required, one for the initContainer, which is already requiring privileged permissions and root access, and one for the es container, which doesn't require the same level.

jordansissel:
I'm not following, though there are gaps in my kubernetes knowledge. Do you mean PodSecurityPolicy, instead of service account?

rwittrick:
A pod security policy is a cluster level resource.  They can be assigned to pods via (cluster) roles and bindings.  One way to add a role to a pod is through service accounts.  Hopefully most cluster admins have implemented the restriced psp linked above and applied it through a CRB to the default service account (like a default least privilege thing)

Every namespace has a `default`  service account that is, by default, applied to every pod if a service account is not supplied in the manifest .

Services that use the `default` namespace service account, like the elasticsearch chart, will auto-inherit the psp assigned to the default service account.

Allowing this chart to set its own service account allows for granular control of security policies throughout a cluster.

Crazybus:
It isn't currently possible to specify serviceAccounts on a container level https://github.com/kubernetes/kubernetes/issues/66020 is the issue tracking adding this in to Kubernetes. 

#38 is the pull request where this was added for Kibana. The only difference being that we will also want to allow the `securityContext` to be set on a pod level (like in the Kibana PR) and for both of the containers. 

@jordansissel https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container might clear it up for you a bit.

jmlrt:
This was implemented in https://github.com/elastic/helm-charts/pull/265, so I'm closing this issue.

PR #333, opened 2019-10-15
[logstash] first version of logstash helm chart

jmlrt:
This is the first version of logstash helm chart.

Related to #69 

## TODO
- [x] Need to add ~`templates/tests`~ & `NOTES.txt`
- [x] Need to add python tests
- [x] Need to add examples with [goss](https://goss.rocks) tests
- [x] Need to add CI jobs
- [ ] Need to optimize resources usage for logstash (persistent disk size, resources requests/limits, java heap)
- [ ] Need to test beats input
- [x] Need to add service ~and ingress for logstash inputs~


logstash/templates/service.yaml
```diff
@@ -0,0 +1,27 @@
...
+  selector:
+    app: "{{ template "fullname" . }}"
+    chart: "{{ .Chart.Name }}"
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+  ports:
+  - name: http
+    port: {{ .Values.service.port }}
+    protocol: TCP
+    targetPort: {{ .Values.httpPort }}
```
jordansissel: I don't understand exposing Logstash's HTTP API (stats, log level changes, etc) as a service with potentially multiple backends. Can you tell me a story about how this might get used?

logstash/values.yaml
```diff
@@ -0,0 +1,219 @@
...
+
+volumeClaimTemplate:
+  accessModes: [ "ReadWriteOnce" ]
+  resources:
+    requests:
+      storage: 1Gi
+
+rbac:
+  create: false
+  serviceAccountName: ""
```
jordansissel: Curious, does Logstash need a kubernetes service account for some purpose?

logstash/values.yaml
```diff
@@ -0,0 +1,219 @@
...
+# This is the node affinity settings as defined in
+# https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity-beta-feature
+nodeAffinity: {}
+
+# The default is to deploy all pods serially. By setting this to parallel all pods are started at
+# the same time when bootstrapping the cluster
+podManagementPolicy: "Parallel"
+
+protocol: http
+httpPort: 9600
```
jordansissel: I don't recommend exposing Logstash's http API as it is, to my knowledge, unauthenticated and has an API for mutation of things like log level.

logstash/values.yaml
```diff
@@ -0,0 +1,219 @@
...
+  initialDelaySeconds: 30
+  periodSeconds: 10
+  timeoutSeconds: 5
+  failureThreshold: 3
+  successThreshold: 1
+
+readinessProbe:
+  httpGet:
+    path: /
+    port: http
```
jordansissel: readiness is complicated. A successful HTTP call to port 9600 doesn't necessarily mean Logstash is capable of serving business needs.

I don't have any alternative solutions in mind, but I'll try to think of something. In the meantime, this comment is not a blocker ;)

logstash/values.yaml
```diff
@@ -0,0 +1,219 @@
...
+##
+schedulerName: ""
+
+nodeSelector: {}
+tolerations: []
+
+# Enabling this will publically expose your Logstash instance.
+# Only enable this if you have security enabled on your cluster
+ingress:
+  enabled: false
```
jordansissel: Similar comment as above. Exposing Logstash's monitoring/management API via ingress doesn't seem safe.

I'd maybe want an ingress (or just clusterip loadbalancer service) if I had something like an `http` input, but I personally would not expose the monitoring/management API endpoint.

logstash/values.yaml
jmlrt: Also a leftover of Elasticsearch copy/paste. I don't know yet if we have a specific need for Logstash yet, but all other charts are managing ServiceAccount

jmlrt: So while Logstash may not have specific need of a ServiceAccount, it seems to be really usefull for people implementing [Pod Security Policies](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) in their K8S cluster (see https://github.com/elastic/helm-charts/issues/61#issuecomment-463784565).

logstash/values.yaml
jmlrt: Service and Ingress for Logstash HTTP API where removed in https://github.com/elastic/helm-charts/commit/1768d4b24c4130566ff47017013bbb7ff8d1fde2.

Do you think we should also remove port exposition at [Pod level](https://github.com/elastic/helm-charts/pull/333/files#diff-e1831a0b9e9999a7d2f6f2276f9fd95aR143-R145)?

logstash/values.yaml
jmlrt: Yes I guess I'll start like that and we'll see how we can set a better probe in a second time.

logstash/values.yaml
jmlrt: Removed in https://github.com/elastic/helm-charts/commit/1768d4b24c4130566ff47017013bbb7ff8d1fde2.

logstash/examples/default/values.yaml
```diff
@@ -0,0 +1,17 @@
+extraEnvs:
+  - name: XPACK_MONITORING_ENABLED
+    value: "false"
+
+#logstashConfig:
+#  logstash.yml: |
+#    xpack.monitoring.enabled: false
```
jmlrt: Note that commenting `XPACK_MONITORING_ENABLED` environment variable and uncommenting `logstash.yml` config is currently failing.

```
cat << EOF > values.yaml
#extraEnvs:
#  - name: XPACK_MONITORING_ENABLED
#    value: "false"
logstashConfig:
  logstash.yml: |
    xpack.monitoring.enabled: false
EOF

# this is failing
helm install --values values.yaml logstash

kubectl logs logstash-logstash-0
2019/10/16 18:40:55 Setting 'node.name' from environment.
2019/10/16 18:40:55 error: open /usr/share/logstash/config/logstash.yml: read-only file system
```

This seem related to [env2yaml](https://github.com/elastic/logstash/blob/master/docker/data/logstash/env2yaml/env2yaml.go) trying to edit the mounted `/usr/share/logstash/config/logstash.yml` to add `node.name` environment variable that I set in [container spec](https://github.com/elastic/helm-charts/pull/333/files#diff-e1831a0b9e9999a7d2f6f2276f9fd95aR149-R152). This part comes from Elasticsearch chart copy/paste so I guess I should remove it, however AFAIK [logstashconfig Volume](https://github.com/elastic/helm-charts/pull/333/files#diff-e1831a0b9e9999a7d2f6f2276f9fd95aR171-R173) shouldn't be mounted read-only, so I don't know why it's not writable by `env2yaml`.

logstash/templates/statefulset.yaml
```diff
@@ -0,0 +1,182 @@
...
+{{ toYaml .Values.livenessProbe | indent 10 }}
+        readinessProbe:
+{{ toYaml .Values.readinessProbe | indent 10 }}
+        ports:
+        - name: http
+          containerPort: {{ .Values.httpPort }}
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
+        env:
+          - name: node.name
```
jmlrt: I need to remove this as told in [previous comment](https://github.com/elastic/helm-charts/pull/333/files#r335650733).

logstash/templates/statefulset.yaml
```diff
@@ -0,0 +1,182 @@
...
+          {{- end }}
+          {{- range .Values.secretMounts }}
+          - name: {{ .name }}
+            mountPath: {{ .path }}
+            {{- if .subPath }}
+            subPath: {{ .subPath }}
+            {{- end }}
+          {{- end }}
+          {{- range $path, $config := .Values.logstashConfig }}
+          - name: logstashconfig
```
jmlrt: As we can specify more than one file in `logstashConfig` value, the name shouldn't be hardcoded.

logstash/examples/default/values.yaml
jmlrt: So as configmap are bind-mounted, it seems they can't be edited even if they are seen as read-write in the pod description:

```
kubectl describe pod helm-logstash-default-logstash-0
...
Containers:
  logstash:
...
    Mounts:
      /usr/share/logstash/config/logstash.yml from logstashconfig (rw,path="logstash.yml")
...
Volumes:
...
  logstashconfig:
    Type:      ConfigMap (a volume populated by a ConfigMap)
    Name:      helm-logstash-default-logstash-config
    Optional:  false
...
```

In summary, you can't mix configuration using configmap and environment variables for `logstash.yml` (details in https://github.com/elastic/helm-charts/pull/333/commits/bcb7af6eb128a54c1c3e61124ff6a1f0b0709aaa).

jmlrt:
[Chart Tests](https://helm.sh/docs/chart_tests/#chart-tests) are not relevant for Logstash as we don't use service while chart tests usually imply running another pod which request some service resource.

logstash/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Requirements
+
+* [Helm](https://helm.sh/) >= 2.8.0
+* Kubernetes >= 1.8
+
+## Usage notes and getting started
+
+* This repo includes a number of [example](./examples) configurations which can be used as a reference. They are also used in the automated testing of this chart
+* Automated testing of this chart is currently only run against GKE (Google Kubernetes Engine). If you are using a different Kubernetes provider you will likely need to adjust the `storageClassName` in the `volumeClaimTemplate`
```
Crazybus: I know this is copy pasted from the Elasticsearch chart but this isn't true anymore. The chart used to default to literally `default` now its not specified so it will use whatever your actual default storage class is. I think we can just remove this sentence for this chart and Elasticsearch.

logstash/examples/default/values.yaml
```diff
@@ -0,0 +1,14 @@
+logstashConfig:
```
Crazybus: This is probably in here just for your testing purposes. However it would be nice to keep the "default" example to only use the default values (so no values). This example mostly exists to make sure that the default install results in a working and healthy deployment. For logstash its a bit weird since it won't be doing anything by default. However it is still nice to have a test making sure that `helm install --name logstash elastic/logstash` doesn't result in logstash failing to startup.

logstash/templates/configmap-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    app: "{{ template "fullname" . }}"
+    chart: "{{ .Chart.Name }}"
+    heritage: {{ .Release.Service | quote }}
+    release: {{ .Release.Name | quote }}
+data:
+{{- range $path, $config := .Values.logstashConfig }}
+  {{ $path }}: |
+{{- if (eq $path "logstash.yml") }}
+    node.name: {{ template "fullname" $ }}
```
Crazybus: This is going to be the release name which won't be unique per replica in the statefulset. 

Looking at https://www.elastic.co/guide/en/logstash/current/logstash-settings-file.html it just defaults to the hostname so I think we can just leave it out. 

If it isn't auto detecting things for some reason I think we could do something like https://github.com/elastic/helm-charts/blob/340868c54c44ae97fd9ef8fa6d682472098dc816/elasticsearch/templates/statefulset.yaml#L334-L338

But instead map it to a different environment variable that Logstash doesn't run env2yaml on and refer to it with something like:

`node.name $OTHER_VAR_NAME` in the default config

logstash/templates/statefulset.yaml
```diff
@@ -0,0 +1,184 @@
...
+    heritage: {{ .Release.Service | quote }}
+    release: {{ .Release.Name | quote }}
+    {{- range $key, $value := .Values.labels }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      app: "{{ template "fullname" . }}"
+  replicas: {{ default .Values.replicas }}
```
Crazybus: So this is in the Elasticsearch chart too so probably by fault. But I don't think this `default` actually does anything. It's always set by default and there is no other value for if its not set.

logstash/templates/statefulset.yaml
```diff
@@ -0,0 +1,184 @@
...
+        {{- range $key, $value := .Values.podAnnotations }}
+        {{ $key }}: {{ $value | quote }}
+        {{- end }}
+        {{/* This forces a restart if the configmap has changed */}}
+        {{- if .Values.logstashConfig }}
+        configchecksum: {{ include (print .Template.BasePath "/configmap-config.yaml") . | sha256sum | trunc 63 }}
+        {{- end }}
+        {{/* This forces a restart if the configmap has changed */}}
+        {{- if .Values.logstashPipeline }}
+        configchecksum: {{ include (print .Template.BasePath "/configmap-pipeline.yaml") . | sha256sum | trunc 63 }}
```
Crazybus: This one should have a different annotation, otherwise Kubernetes will ignore the first `configchecksum` if a config and a pipeline are configured.

logstash/templates/statefulset.yaml
```diff
@@ -0,0 +1,184 @@
...
+{{ toYaml .Values.livenessProbe | indent 10 }}
+        readinessProbe:
+{{ toYaml .Values.readinessProbe | indent 10 }}
+        ports:
+        - name: http
+          containerPort: {{ .Values.httpPort }}
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
+        env:
+{{- if not ".Values.logstashConfig.logstash.yml" }}
```
Crazybus: I commented on this for the configmap already. But just noting it here as well, do we need to set this or does the default hostname work?

logstash/values.yaml
```diff
@@ -0,0 +1,197 @@
...
+# Extra environment variables to append to this nodeGroup
+# This will be appended to the current 'env:' key. You can use any of the kubernetes env
+# syntax here
+extraEnvs: []
+#  - name: MY_ENVIRONMENT_VAR
+#    value: the_value_goes_here
+
+# A list of secrets and their paths to mount inside the pod
+secretMounts: []
+
```
Crazybus: I think leaving in the commented example like the Elasticsearch chart would make it easier for users to know the formatting.

logstash/values.yaml
```diff
@@ -0,0 +1,197 @@
...
+imagePullSecrets: []
+
+podAnnotations: {}
+
+# additionals labels
+labels: {}
+
+logstashJavaOpts: "-Xmx1g -Xms1g"
+
+resources:
```
Crazybus: I think these are just copied for Elasticsearch for now. But we should chat with the Logstash team (or check docs) to see what the recommended defaults would be.

logstash/README.md
jmlrt: fixed in https://github.com/elastic/helm-charts/pull/333/commits/ef35da4343010cdcd4112d250009bde97f7f0e8c (for Logstash) and in https://github.com/elastic/helm-charts/pull/335 (for Elasticsearch).

logstash/examples/default/values.yaml
jmlrt: Fixed in https://github.com/elastic/helm-charts/pull/333/commits/e2f43469596af1fa50788eaff236f38f29318a1c

logstash/values.yaml
jmlrt: Yeah for this one I'd like to add a commented example. Do you have any good use case needing a secret mount in Logstash so we can have a relevant example.

logstash/templates/configmap-config.yaml
jmlrt: AFAIK, `node.name` using `hostname` as default value should be OK.
Fixed in https://github.com/elastic/helm-charts/pull/333/commits/cad113038b875e0d1db2dd548a26dc6cb190bd85.

logstash/templates/statefulset.yaml
jmlrt: Default hostname should be OK, Fixed in https://github.com/elastic/helm-charts/pull/333/commits/cad113038b875e0d1db2dd548a26dc6cb190bd85.

logstash/templates/statefulset.yaml
jmlrt: fixed in https://github.com/elastic/helm-charts/pull/333/commits/4af4a4b4895dcde79a33661e828cda7115d7a49a

jmlrt:
I guess this is ready for a second review and hope to be able to merge it and add the missing parts in following PR.

Status of what's not fully completed in this PR:

- [ ] Need to optimize resources usage for logstash (persistent disk size, resources requests/limits, java heap)

    I didn't find any recommendation for default values in [Logstash documentation](https://www.elastic.co/guide/en/logstash/current/index.html). I guess we can add them in a following PR.

- [ ] Need to test Beats input

    Beats input has been tested using services provided in https://github.com/elastic/helm-charts/pull/333/commits/68a85c0ab1e53179d26bcd7590414d40cd92f86c and the [Filebeat chart](https://github.com/elastic/helm-charts/tree/master/filebeat) locally.

    We have 3 solution to add an automated test:

    - Reuse some Beats coming from [Filebeat examples](https://github.com/elastic/helm-charts/tree/master/filebeat/examples) in the same way we are reusing Elasticsearch coming from [Elasticsearch examples](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples) in other chart tests.
    In this case should Logstash and Filebeat test run in // or do we need to start one before the other?

    - Do not reuse Beats from another test but a new one (using `helm install elastic/filebeat`) in the `make install` target of the new `logstash/examples/beats` test

    - Adding a new "full stack test" outside of the examples of each charts which would deploy and tests all the chart together (filebeat + metricbeat + elasticsearch + logstash + kibana) so we don't need to add a new test for all the combinations of products integration we could have inside specific charts examples. This way also ensure that all chart are running in the same time, while we can't be 100% sure of the order in different sub jobs of a matrix job.

- [ ] Need to add service ~and ingress for logstash inputs~

    Service has been added in https://github.com/elastic/helm-charts/pull/333/commits/68a85c0ab1e53179d26bcd7590414d40cd92f86c, I'm not sure if we really want to manage Ingress for the "listener inputs". I guess if we need that we can add it in a following PR.

cc @Crazybus @jordansissel

logstash/values.yaml
Crazybus: Best example would be for mounting certificates to be able to talk to Elasticsearch with security enabled.

logstash/values.yaml
```diff
@@ -0,0 +1,199 @@
...
+  #     command: ["/bin/sh", "-c", "echo Hello from the postStart handler > /usr/share/message"]
+  # postStart:
+  #   exec:
+  #     command: ["/bin/sh", "-c", "echo Hello from the postStart handler > /usr/share/message"]
+
+service: {}
+#  annotations: {}
+#  type: ClusterIP
+#  ports:
+#    - name: beats
```
Crazybus: Nice! It's really cool that you can just support multiple ports and types this way.

masterkain:
can't see this on the repo yet

```
NAME                 	CHART VERSION	APP VERSION	DESCRIPTION
elastic/elasticsearch	7.4.1        	7.4.1      	Official Elastic helm chart for Elasticsearch
elastic/filebeat     	7.4.1        	7.4.1      	Official Elastic helm chart for Filebeat
elastic/kibana       	7.4.1        	7.4.1      	Official Elastic helm chart for Kibana
elastic/metricbeat   	7.4.1        	7.4.1      	Official Elastic helm chart for Metricbeat
```

wanted to give it a try, thanks!

jmlrt:
Hi @masterkain,

Logtsash chart hasn't been uploaded to Elastic Helm repository yet. It should be released soon on Helm repos with 7.5.0 release.

In the meantime, you need to install it from GitHub repo:

```
git clone https://github.com/elastic/helm-charts.git
cd helm-chart/logstash && helm install --name logstash .
```

PR #371, opened 2019-11-14
Update values.yaml

dwaiba:

### DIFF
--- a/filebeat/templates/daemonset.yaml
+++ b/filebeat/templates/daemonset.yaml
@@ -67,7 +67,7 @@ spec:
           type: DirectoryOrCreate
       - name: varlibdockercontainers
         hostPath:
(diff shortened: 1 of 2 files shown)


fatmcgav:
@dwaiba Thankyou for opening a PR...

Are you able to elaborate on what problem you're trying to fix with these updates?

dwaiba:
The log paths for filebeat not work for 1.14.3 since log paths for k8s 1.14.3 and onwards have changed. Also made the autodiscover default

fatmcgav:
@dwaiba Are you able to confirm whether you're using a cloud Kubernetes platform, or if you're running on a self-hosted bare metal setup? 

As `/data/shared/docker` isn't a path structure we've seen used in any of the cloud K8s platforms...

Issue #376, opened 2019-11-18
Control plane availability impacts data plane availability


erik-stephens:
**Chart version:**

7.2.0

**Kubernetes version:**

1.15

**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)

on-prem

**Helm Version:**

2.15

**Describe the bug:**

Pods should remain ready even though masters not responding in a timely manner.

**Steps to reproduce:**

1. Create a 3-node master-only StatefulSet and a 1-node data-only StatefulSet.  Calling the cluster "foo":
1. `kubectl scale statefulset foo-master --replicas=1`
1. Wait for data node to become unready: `kubectl get pod foo-data-0 -w`
1. `kubectl get service foo-data | grep Endpoints`
1. `kubectl exec foo-data-0 -- curl -s --fail http://localhost:9200/`

**Expected behavior:**

Pod to remain in ready state, perhaps by using `/_cluster/health?local=true` instead of `/`.

**Provide logs and/or server output (if relevant):**
```bash
# k describe pod t1-data-0 | tail -n1
  Warning  Unhealthy  33s (x90 over 30m)  kubelet, r7c6n3  Readiness probe failed: Elasticsearch is already running, lets check the node is healthy

# k exec t1-data-0 -- curl -s --fail http://localhost:9200/
command terminated with exit code 22

# k exec t1-data-0 -- curl -s --fail http://localhost:9200/_cluster/health?local=true | jq .
{
  "cluster_name": "t1",
  "status": "red",
  "timed_out": false,
  "number_of_nodes": 3,
  "number_of_data_nodes": 1,
  "active_primary_shards": 0,
  "active_shards": 0,
  "relocating_shards": 0,
  "initializing_shards": 0,
  "unassigned_shards": 0,
  "delayed_unassigned_shards": 0,
  "number_of_pending_tasks": 0,
  "number_of_in_flight_fetch": 0,
  "task_max_waiting_in_queue_millis": 0,
  "active_shards_percent_as_number": "NaN"
}
```

**Any additional context:**

Elasticsearch 6.8.1

https://github.com/elastic/helm-charts/blob/0411010e8f03c25c06d864ce62f5b482b5439bdb/elasticsearch/templates/statefulset.yaml#L215

[...]

PR #380, opened 2019-11-20
[elasticsearch] Tweak the 'readinessProbe' command to verify that master nodes are available.

fatmcgav:
The behaviour of the `/` endpoint changed[0] between 6.x and 7.x, whereby
previously it would return a HTTP `503` response when the cluster was
blocked, it now returns a HTTP `200` response even if there are no masters
available.

This change updates the behaviour of the `readinessProbe` command during
normal running to verify that the local node is responding and that there
are master nodes available. [1]

The desired behaviour here is that if the data nodes are unable to talk to
their master nodes for whatever reason, then the data nodes will become
`Unready` and therefore be removed from the Service load-balancer until
the master nodes are available again.

Refs:
[0] https://github.com/elastic/elasticsearch/pull/29045
[1] https://github.com/elastic/elasticsearch/issues/34897#issuecomment-519441101

- ~[ ] README.md updated with any new values or changes~
- ~[ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`~

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -211,8 +211,8 @@ spec:
                 }
 
                 if [ -f "${START_FILE}" ]; then
-                    echo 'Elasticsearch is already running, lets check the node is healthy'
-                    http "/"
+                    echo 'Elasticsearch is already running, lets check the node is healthy and there are master nodes available'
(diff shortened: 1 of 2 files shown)


fatmcgav:
Related issue - https://github.com/elastic/helm-charts/issues/376

PR #398, opened 2019-12-06
[logstash] Add FAQ: Why am I missing required field `serviceName` during install?

drenze:
User may receive the following error when executing `helm install ...`:

Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec

This is due to the requirement to define at least one service in
`values.yaml` prior to an install. Added FAQ to README.md to address
this question.

### DIFF
--- a/logstash/README.md
+++ b/logstash/README.md
@@ -137,6 +137,44 @@ There are a couple reasons we recommend this.
 1. Tying the availability of Logstash to the download service to install plugins is not a great idea or something that we recommend. Especially in Kubernetes where it is normal and expected for a container to be moved to another host at random times.
 2. Mutating the state of a running docker image (by installing plugins) goes against best practices of containers and immutable infrastructure.
 
+#### Why am I missing required field `serviceName` during install?
+
+When executing `helm install`, you may receive an error similar to the following:
+
+```bash
+Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec
+```
+
+By default, no services are defined in the `values.yaml` file. This is due to the fact that there are many plugins and many possible use cases for Logstash.
+
+In order to successfully install logstash, at least one service must be defined. This can be done by finding the following section in `values.yaml`:
+
+```yaml
+service: {}
+#  annotations: {}
+#  type: ClusterIP
+#  ports:
+#    - name: beats
+#      port: 5044
+#      protocol: TCP
+#      targetPort: 5044
+#    - name: http
+#      port: 8080
+#      protocol: TCP
+#      targetPort: 8080
+```
+
+Once you have found this, do the following:
+
+1. Remove `{}` from `service:`.
+2. Uncomment the following lines:
+   ```yaml
+   #  annotations: {}
+   #  type: ClusterIP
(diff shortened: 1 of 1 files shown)


jmlrt:
Hi @drenze,
Thank you for your PR, however service are optional and used only if you need to define a Logstash listener to receive external calls like the [Beats input plugin](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-beats.html) which will receive logs from Beats on port 5044 for exemple.

If you don't need a service, you should let the default config with `service: {}`. This configuration is automatically tested everyday without any issue.

In order to help you with your issue, can you paste here your values file and the exact command you are typing?

drenze:
Thanks for the response. 

I received that error when attempting to deploy the unmodified chart, with an unmodified values file. Steps to reproduce are to clone the repo, then install the chart. 

Douglas J. Renze
drenze@gmail.com

"Unless you are willing to do the ridiculous, God will not do the miraculous."

> - Mother Angelica


________________________________

jmlrt:
We have a python [unit test](https://github.com/elastic/helm-charts/blob/master/logstash/tests/logstash_test.py#L92-L94) which specifically ensure than no service will be required with default configuration. This test is run every day on our [CI](https://devops-ci.elastic.co/job/elastic+helm-charts+master+template-testing/CHART=logstash,label=docker&&virtual/303/console).

In addition we have another daily [CI job](https://devops-ci.elastic.co/job/elastic+helm-charts+master+integration-logstash/KUBERNETES_VERSION=1.14,LOGSTASH_SUITE=default,label=docker&&virtual/) which is deploying helm chart with default values (by cloning repo then installing the chart) on a GKE cluster and run [smoke tests](https://github.com/elastic/helm-charts/blob/master/logstash/examples/default/test/goss.yaml) to ensure that the chart installation is OK and Logstash is working well.

Both this tests are running without error.

Can you provide the exact commands you are typing and some details on your k8s environment:
- K8S provider (GKE, AKS, EKS, OpenShift, bare metal, minikube, ...)
- K8S version
- Helm version

krichter722:
This seems to happen with helm 3.x (3.0.1 in my case) only. A link to the issue or roadmap for support of helm 3.x would be welcome.

jmlrt:
OK, we don't support Helm v3 yet as mentionned in [elastic/helm-charts#helm-versions](https://github.com/elastic/helm-charts#helm-versions). Despite we will support it at some time, we don't have any roadmap for it currently.

krichter722:
The workaround proposed in this PR seems to work. I opened #401 not to push it, but it has to be done one day and it's a simple first step.

jmlrt:
I'm closing this PR as it is explicitely mentionned than Helm v3 is not supported, and we now have an issue to support it.

poochwashere:
I am getting this same error with `helm 2.14`
```
❯ helm --tls --tiller-namespace=tiller --namespace=elk install --name logstash elastic/logstash --values values.yaml
Error: validation failed: error validating "": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec
```

jmlrt:
@poochwashere can you provide your `values.yaml`

PR #406, opened 2019-12-11
Prefixed helper functions with chart name

bpdunni:
It is a best practice to prefix helm template helper functions with the chart name. Since template functions are global, this can lead to bugs because the last function to be loaded will be used. The best solution is to prefix the template function with the chart name.

See [Named Tempaltes](https://helm.sh/docs/topics/chart_template_guide/named_templates/)

- [ *] Chart version *not* bumped (the versions are all bumped and released at the same time)


elasticcla:
Hi @bpdunni, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/helm-charts/pull/406.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #407, opened 2019-12-11
Prefixed helper functions with chart name

bpdunni:
It is a best practice to prefix helm template helper functions with the chart name. Since template functions are global, this can lead to bugs because the last function to be loaded will be used. The best solution is to prefix the template function with the chart name.

See [Named Templates](https://helm.sh/docs/topics/chart_template_guide/named_templates/)

- [ *] Chart version *not* bumped (the versions are all bumped and released at the same time)


jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

PR #432, opened 2020-01-05
Add an example, tests and README.md hints for installation of filebeat on a single node cluster implementation

krichter722:

### DIFF
--- a/filebeat/examples/microk8s/Makefile
+++ b/filebeat/examples/microk8s/Makefile
@@ -1,0 +1,13 @@
+default: test
+
+include ../../../helpers/examples.mk
+
+RELEASE := helm-filebeat-default
+
+install:
+	helm upgrade --wait --timeout=600 --install $(RELEASE) --values values.yaml ../../
+
+test: install goss
+
+purge:
+	helm del --purge $(RELEASE)
--- a/filebeat/examples/microk8s/test/goss.yaml
+++ b/filebeat/examples/microk8s/test/goss.yaml
@@ -1,0 +1,51 @@
+port:
+  tcp:5066:
+    listening: true
+    ip:
+    - '127.0.0.1'
+
+mount:
+  /usr/share/filebeat/data:
+    exists: true
+  /run/docker.sock:
+    exists: true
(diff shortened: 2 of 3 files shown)


jmlrt:
Hi @krichter722,
Thanks for this PR.

Mounting `/var/log` is not a specific need for single node cluster nodes but could be added to default configuration.

We were planning to add it to default config with #294, however this PR is stale and `/var/data` shouldn't be part of default config.

If you want to open a PR with only `/var/log` part of #294 instead, that could be great.

krichter722:
@jmlrt Since #294 is closed by the stale bot, but has a good looking patch, I suggest we should work on that. You can close this PR and mention me in #294 if you need any support.

Btw, the stale bot is insanely annoying and causes way more effort than it saves, please get rid of it.

jmlrt:
Hi @krichter722 `/var/log` mount has been added to the Filebeat pod in #568.

Now that it is part of the default chart. I don't see any difference between the configuration of your microk8s example and the default configuration.

[default example](https://github.com/elastic/helm-charts/blob/master/filebeat/examples/default/README.md) should work without any changes on microk8s and a specific example for it doesn't seem required.

I'll close this PR if it's fine for you.

> Btw, the stale bot is insanely annoying and causes way more effort than it saves, please get rid of it.

Using bots marking GitHub issues and PR as stale or closing them after a defined period is a common best practice for maintainers and is used by the majority of GitHub repositories (ie: Kubernetes, Helm, ...).

I understand it's not user friendly for you to post a "still valid" comment every 3 months when you have a few issues opened that we don't have time to handle.

On the other hand, as a maintainer having to deal with tenth and tenth of issues/PR on different repositories, I find it almost mandatory to avoid being overwelmed by tenth and tenth of issues and PR which can be no more relevant or duplicates or where we don't have any answers from the creator...

krichter722:
@jmlrt Thanks for your feedback. Closing the PR is a good idea. I'm happy about the changes to filebeat.

About stale bot: I get it, however you almost might not have been able to list helm as an example project using it as I was close to getting the devs to drop stale bot for the helm project :)

PR #445, opened 2020-01-14
dont set default resource values

### LINKED ISSUE #409: [Elasticsearch] Default values for resource policies makes it hard to have unset limits
**Describe the feature:** Set empty object as default value for resource policy for elasticsearch containers

**Describe a specific use case for the feature:** Often it is quite useful to have applications running without limits in its resource policy (for example, setting CPU limits causes throttling CPU time). At the moment is not easy to unset default values, forcing the user to set limits.

pbecotte:
- [X] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

This should close #409


jmlrt:
As mentionned in https://github.com/elastic/helm-charts/issues/409#issuecomment-577932767 we'll close this PR and the issue related.

Issue #446, opened 2020-01-14
[metricbeat] Different tolerations, nodeSelector and affinity for deployment and daemonset


FedeBev:
**Describe the feature:**
In my opinion would be great to have a different value option for tolerations, nodeSelector and affinity of daemonset and deployment.

An example of value.yaml could be the following:
```yaml
daemonset:
  nodeSelector: {}

  tolerations:
    - operator: "Exists"

deployment:
  nodeSelector:
    my_key: "true"

  tolerations:
    - key: my_key
      value: "true"
      operator: Equal
      effect: NoExecute
```

**Describe a specific use case for the feature:**

In our case, for example, we have a node pool dedicated to elastic stack and we would like to keep all the pods on it, when possible. This is for billing reason and to prevent the infrastructure components load from disturbing customer applications

If PR are welcome I can deal with that

mitchellmaler:
Just ran into this same issue. We have dedicated nodes that have specific resource requirements but the metricbeat deployment is deploying to those nodes.

jmlrt:
Hi @FedeBev, we have #448 open for that

PR #448, opened 2020-01-15
split some of the deployment and daemonset config to make things more…

pbecotte:
… flexible

Looking at #446 realized we may want to split daemonset/deployment config in some places. This is kind of a big change to the values file though, so not sure if theres a better way to transition. Consider this as a proposal :)


FedeBev:
I agree with you: this is a big change. However I can't a better way to do that.

Thanks

jmlrt:
jenkins test this please

mitchellmaler:
Can this PR be merged in? We are still having issues with the resources being the same on all pods even though the daemonset pods use much less compared to the deployment pod.

jmlrt:
Hey @pbecotte,
Thanks for your PR, however, this is a breaking change and we'd like to ensure user with old values can still use the chart after splitting config, so I started a rework in #572, which should allow re-using old values if they are still used.

jmlrt:
Closed by #572 
Thanks for your work @pbecotte

Issue #394, opened 2019-12-06
[Metricbeat] DNS lookup failure for node host


4nte:
**Chart version:** 7.4.1

**Kubernetes version:** 1.14.2

**Kubernetes provider:** Digital Ocean

**Helm Version:** 2.13

**`helm get release` output**
```
REVISION: 8
RELEASED: Fri Dec  6 12:57:01 2019
CHART: metricbeat-7.4.1
USER-SUPPLIED VALUES:
metricbeatConfig:
  kube-state-metrics-metricbeat.yml: |-
    metricbeat.modules:
    - module: kubernetes
      enabled: true
      metricsets:
        - state_node
        - state_deployment
        - state_replicaset
        - state_pod
        - state_container
      period: 10s
      hosts: ["${KUBE_STATE_METRICS_HOSTS}"]
    output.elasticsearch:
      hosts: '${ELASTICSEARCH_HOSTS:elasticsearch-master:9200}'
  metricbeat.yml: |
    system:
      hostfs: /hostfs
    metricbeat.modules:
    - module: kubernetes
      metricsets:
        - container
        - node
        - pod
        - system
        - volume
      period: 10s
      host: "${NODE_NAME}"
      hosts: ["${NODE_NAME}:10255"]
      processors:
      - add_kubernetes_metadata:
          in_cluster: true
    - module: kubernetes
      enabled: true
      metricsets:
        - event
    - module: system
      period: 10s
      metricsets:
        - cpu
        - load
        - memory
        - network
        - process
        - process_summary
      processes: ['.*']
      process.include_top_n:
        by_cpu: 5
        by_memory: 5
    - module: system
      period: 1m
      metricsets:
        - filesystem
        - fsstat
      processors:
      - drop_event.when.regexp:
          system.filesystem.mount_point: '^/(sys|cgroup|proc|dev|etc|host|lib)($|/)'
    - module: golang
      metricsets:
        - expvar
        - heap
      period: 10s
      hosts: ["${NODE_NAME:6060}"]
      heap.path: "/debug/vars"
      expvar:
        namespace: "example"
        path: "/debug/vars"
    output.elasticsearch:
      hosts: '${ELASTICSEARCH_HOSTS:elasticsearch-master:9200}'

COMPUTED VALUES:
affinity: {}
clusterRoleRules:
- apiGroups:
  - extensions
  - apps
  - ""
  resources:
  - namespaces
  - pods
  - events
  - deployments
  - nodes
  - replicasets
[...]

PR #449, opened 2020-01-17
Fix DNS lookup failure error for host URLs

FedeBev:
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml` (Seems to be not required for current tests)

This is a fix PR for #394 bug, feel free to suggest any other solution/idea.

Thanks


FedeBev:
CLA has now been signed.
Sorry, what I have to do now?

pbecotte:
I wound up fixing this like this-

https://github.com/elastic/helm-charts/pull/439#issuecomment-573964098

FedeBev:
> I wound up fixing this like this-
> 
> [#439 (comment)](https://github.com/elastic/helm-charts/pull/439#issuecomment-573964098)

That could be a great solution!
Both solutions are valid and feasible in my opinion, maybe I cane make this options configurable through values file, what do you think?

pbecotte:
I had an earlier PR that set this variable, you could resurrect it and trim it down to just this. Eventually I came to the conclusion that host networking was the easiest solution, but almost never the right one. That's not up to me though ;)

jmlrt:
Hi @FedeBev, 
This was already discussed a lot, host networking can be a great addition to this chart as long as it is optional and disabled by default.
You can also find an example in @pbecotte's [Filebeat PR](https://github.com/elastic/helm-charts/pull/321) (this was disabled by default in https://github.com/elastic/helm-charts/pull/391/commits/538af31822d3bbbac93b1a270493b4818a983666).

FedeBev:
Hi @jmlrt ,
`hostNetwork` and `dnsPolicy` options are now configurable through values.

In my opinion also @pbecotte solution is fine, but I can't see the new options in 7.6 release and in my opinion is needful.

Please, could you let me know the state on that?

Thanks

jmlrt:
@FedeBev I created #585 to add host networking to Metricbeat while staying consistent with Filebeat.
This will be merged today or in a couple of days so I'll close your PR.

I'm sorry for making you wait so long then replacing your work due to bad timing.
I hope it won't demotivate you for contributing to this repos in the future.

FedeBev:
Hi @jmlrt, thank's for you work.

I kept the options separate to leave maximum configuration flexibility. At the moment I can't image a situation where this is really required, so it's perfectly fine for me to keep that configuration in one option and #585 is perfect for my use case.

Don't worry for the timing,  luckily the workaround was very easy 😄

PR #458, opened 2020-01-23
[elasticsearch] set cpu request = cpu limit

jmlrt:

Related to https://github.com/elastic/helm-charts/issues/409#issuecomment-577583888

I think we can let cpu request to 100m in "workstation" examples like [docker-for-mac example](https://github.com/elastic/helm-charts/blob/b2d5d8640c/elasticsearch/examples/docker-for-mac/values.yaml#L11).

What about default values for [Filebeat](https://github.com/elastic/helm-charts/blob/b2d5d8640c/filebeat/values.yaml#L71-L76), [Metricbeat](https://github.com/elastic/helm-charts/blob/b2d5d8640c/metricbeat/values.yaml#L141-L146), [Kibana](https://github.com/elastic/helm-charts/blob/b2d5d8640c/kibana/values.yaml#L35-L40) and [Logstash](https://github.com/elastic/helm-charts/blob/b2d5d8640c/logstash/values.yaml#L48) should we also stay with the request=limit rule?


jmlrt:
jenkins test this please

jmlrt:
Upgrade test is failing when changing the cpu request: https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+integration-elasticsearch/254/ES_SUITE=upgrade,KUBERNETES_VERSION=1.13,label=docker&&virtual/

I'll investigate next week.

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

helpers/terraform/variables.tf
```diff
@@ -39,7 +39,7 @@ variable "additional_zones" {
 
 variable "machine_type" {
   description = "Machine type for the kubernetes nodes"
-  default     = "n1-standard-8"
+  default     = "custom-10-30720"
```
Crazybus: The fact that this increase is causing issues for our testing environment might be a sign that this should be announced as a breaking chance in the release notes. At least breaking for anybody using the current default resource limits.

jmlrt: In our testing environment we are deploying 21 Elasticsearch pods (7 scenarios x 3 nodes) per clusters and we can assume that the pod are never requiring more resources than set in requests resources during the tests.

This change cause resource needed for Elasticsearch pods grow from 2.1 CPU to 21 CPU in our testing environment.

IMHO standard usage is closer to 3 long running Elasticsearch pods per cluster which are consuming CPU and memory limits, so I'm not sure that the impact should be considered as a breaking change. WDYT?

Crazybus: The impact was big enough to cause us issues in our testing environment because of resource capacity. A 10x increase in the amount of CPUs requested by default could also cause issues for others too. For clusters with autoscaling enabled they might suddenly end up with 10 times the amount of Kubernetes node running after a version bump.

I don't think the change is big enough to avoid doing it, but I do think it would be nice to give all users a heads up including documentation for how to keep the existing configuration.

jmlrt: That sounds good to me, I'll merge the PR and will add a note in next release's changelog about this change and how to keep existing config.

jmlrt: @Crazybus notice added in https://github.com/elastic/helm-charts/pull/476/commits/293f0ec38fa00a74fd80c0beee3a67fc2c886f0e

Crazybus: Perfect thank you!

PR #464, opened 2020-01-28
Add wildcard support for ingresses

Matsa59:
Currently ingresses doesn't support wildcard hosts. This PR allow developers to use wildcard for ES. 
We could also update kibana and apm ingresses for tls but this will add breaking changes.


elasticcla:
Hi @Matsa59, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/helm-charts/pull/464.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Issue #363, opened 2019-11-11
[elasticsearch] kubernetes-kind example: init-container doesn't have permission to chown a volume


holms:
**Chart version:**

elasticsearch-7.4.1

**Kubernetes version:**

1.15.5-do.0

**Kubernetes provider:**

Digital Ocean K8s

**Helm Version:**

Not sure, the one which terraform is using in their provider.

**`helm get release` output**

Sorry but this output is already with a fix, been running init-container in privilege mode and it worked.

```
REVISION: 1
RELEASED: Wed Nov  6 17:17:35 2019
CHART: elasticsearch-7.4.1
USER-SUPPLIED VALUES:
esJavaOpts: -Xmx1g -Xms1g
extraInitContainers: |
  - name: create
    image: busybox:1.28
    command: ['mkdir', '-p', '/usr/share/elasticsearch/data/nodes/']
    securityContext:
      runAsUser: 0
    volumeMounts:
     - mountPath: /usr/share/elasticsearch/data
       name: elasticsearch-master
  - name: file-permissions
    image: busybox:1.28
    command: ['chown', '-R', '1000:1000', '/usr/share/elasticsearch/']
    securityContext:
       runAsUser: 0
    volumeMounts:
     - mountPath: /usr/share/elasticsearch/data
       name: elasticsearch-master
nodeSelector:
  doks.digitalocean.com/node-pool: elasticsearch
readinessProbe:
  initialDelaySeconds: 200
resources:
  limits:
    cpu: 1000m
    memory: 2G
  requests:
    cpu: 100m
    memory: 2G
volumeClaimTemplate:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 20G
  storageClassName: do-block-storage

COMPUTED VALUES:
antiAffinity: hard
antiAffinityTopologyKey: kubernetes.io/hostname
clusterHealthCheckParams: wait_for_status=green&timeout=1s
clusterName: elasticsearch
esConfig: {}
esJavaOpts: -Xmx1g -Xms1g
esMajorVersion: ""
extraEnvs: []
extraInitContainers: |
  - name: create
    image: busybox:1.28
    command: ['mkdir', '-p', '/usr/share/elasticsearch/data/nodes/']
    securityContext:
      runAsUser: 0
    volumeMounts:
     - mountPath: /usr/share/elasticsearch/data
       name: elasticsearch-master
  - name: file-permissions
    image: busybox:1.28
[...]

PR #434, opened 2020-01-08
[elasticsearch]  add workaround to fix kind example

jmlrt:
Add workaround to fix kind mount points created from PVCs not writeable by non-root users issue.

FIx #429

PR #472, opened 2020-02-05
access to volume with kind

ryandawsonuk:
Otherwise get AccessDeniedException when running on KIND 0.7.0. See https://github.com/elastic/helm-charts/issues/363

### DIFF
--- a/elasticsearch/examples/kubernetes-kind/values.yaml
+++ b/elasticsearch/examples/kubernetes-kind/values.yaml
@@ -21,3 +21,22 @@ volumeClaimTemplate:
   resources:
     requests:
       storage: 100M
+
+# Access to the volume
+extraInitContainers: |
+   - name: create
+     image: busybox:1.28
+     command: ['mkdir', '-p', '/usr/share/elasticsearch/data/nodes/']
+     securityContext:
+       runAsUser: 0
+     volumeMounts:
+      - mountPath: /usr/share/elasticsearch/data
+        name: elasticsearch-master
+   - name: file-permissions
+     image: busybox:1.28
+     command: ['chown', '-R', '1000:1000', '/usr/share/elasticsearch/']
+     securityContext:
+        runAsUser: 0
+     volumeMounts:
+      - mountPath: /usr/share/elasticsearch/data
+        name: elasticsearch-master


jmlrt:
Hi @ryandawsonuk,
I removed that `extraInitContainer` in #434 because it wasn't needed.

However what is needed when using [KIND](https://github.com/kubernetes-sigs/kind/issues/830) is to install [Rancher Local Path Provisioner](https://github.com/rancher/local-path-provisioner) as mentionned [here](https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/kubernetes-kind/README.md#current-issue).

Did you installed it?

ryandawsonuk:
Yeah I'm running the [local path provisioner](https://github.com/SeldonIO/seldon-core/pull/1369/files#diff-c6c0fdf7cfe6a9667e626244e3414bbeR22). Am wondering if it might be something to do with the new KIND. I found that the old [removed initContainer](https://github.com/elastic/helm-charts/pull/434/files#diff-d992e90f5c32753288bd2a88fdfb2522L23) actually doesn't work either - it has to [set runAsUser](https://github.com/SeldonIO/seldon-core/pull/1369/files#diff-d675dc415576de6cacafcccddd38523dR31) too.

I'm using KIND 0.7.0 to create [a 1.15 cluster](https://github.com/SeldonIO/seldon-core/pull/1369/files#diff-c6c0fdf7cfe6a9667e626244e3414bbeR17). Or could be related to something else in my [KIND config](https://github.com/SeldonIO/seldon-core/pull/1369/files#diff-071817d23aa30e5814e8a7087b7ff2ebR1).

jmlrt:
Hi @ryandawsonuk,
This PR was closed by our bot because it was stale since a month.

As I didn't have time to answer you before I was able to take some time today to try to reproduce with a new Kind version (v0.8.1) and default Kind config.

I couldn't reproduce your `AccessDeniedException` error. Everything is working fine on my test.

> Or could be related to something else in my [KIND config](https://github.com/SeldonIO/seldon-core/pull/1369/files#diff-071817d23aa30e5814e8a7087b7ff2ebR1).

At first sight your config look fine but I'm not an KIND expert.

In addition, it seems that Kind >= 0.7.0 don't need anymore `local-path` storage class and can use `standard` storage class. My tests with `standard` storage class are fine.

I'll create a PR to update `kubernetes-kind` example.

ryandawsonuk:
Cool, thanks for following this up

PR #473, opened 2020-02-06
Add extraContainers and extraInitContainers for all charts

### LINKED ISSUE #327: Support extraContainers (from stable/kibana chart)
Please add the 'extraContainers' feature from the stable/kibana chart
(And also authProxyPort)

We are using this to add a keycloak-proxy sidecar to the kibana pod.

jmlrt:
Should fix #327


jmlrt:
This PR is prettyg big with more than 90% of the changes being in [format python code with black](https://github.com/elastic/helm-charts/pull/473/commits/9029eeadf2989bdc13f9f87555f5f135a25cc485) commit.

Maybe reviewing commit by commit should be easier for this one.

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

jmlrt:
> #327 also mentions porting `authProxyPort` for kibana but it looks like `httpPort` could be used, is that right?

Yes I think `httpPort` should be used for that.

willemm:
I posted the mention about authProxyPort needing to be ported, and I still believe it should be.
After all, if you use any kind of proxy as sidecar of any container, the service.yaml will need to be pointed to the port of the proxy, whereas the main container itself will be listening on a different port.  As it is now, both are configured with httpPort which would never work.  (Well, actually it *will* work, it will just bypass the proxy entirely, which is not what you want).

PR #475, opened 2020-02-10
[meta] format with black python scripts

jmlrt:
This PR is reformatting all python scripts using [black](https://black.readthedocs.io/en/stable/) and add black formating to CI.

Also add [phony targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) and help to Makefiles.

### DIFF
--- a/.ci/jobs/elastic+helm-charts+master+template-lint-python.yml
+++ b/.ci/jobs/elastic+helm-charts+master+template-lint-python.yml
@@ -17,4 +17,6 @@
         #!/usr/local/bin/runbld
         set -euo pipefail
 
+        virtualenv venv && source venv/bin/activate
+        pip install -r requirements.txt
         make lint-python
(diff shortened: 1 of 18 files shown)


.ci/jobs/elastic+helm-charts+master+template-lint-python.yml
```diff
@@ -0,0 +1,20 @@
...
+        type: slave
+        name: label
+        values:
+        - docker&&virtual
+    builders:
+    - shell: |-
+        #!/usr/local/bin/runbld
+        set -euo pipefail
+
+        make lint-python
```
Conky5: Do the python dependencies need installed before this? Otherwise `black` might not be available I think.

jmlrt: nice catch, fixed in https://github.com/elastic/helm-charts/pull/475/commits/8e0c1698a29f64fa5427db6d20ac976ddbb3d50f

jmlrt:
jenkins test this please

PR #485, opened 2020-02-14
Add possibility to define custom readinessProbe

mschmidt291:


fatmcgav:
@mschmidt291 are you able to elaborate on _why_ you need to define a completely custom readiness probe?

The reason I ask is that the existing readiness probe has been written to handle multiple scenarios gracefully, such as initial cluster bootstrapping, rolling restarts, graceful cluster upgrades etc.
By defining a custom readiness probe, a lot of that functionality could end up being broken...

mschmidt291:
@fatmcgav Oh damn, I actually closed the first PR because of the CLA which was missing. I forgot to write the link down here.

There is a bug in Kuberentes with exec readinessProbes, which results in a very high CPU usage. You can read more about it here: https://github.com/kubernetes/kubernetes/issues/82440 We're currently in the process of moving every deployment which we use away from exec to a tcpProbe or httpProbe. This fixes the high CPU usage.

fatmcgav:
@mschmidt291 cheers for the link... That's a pretty painful looking bug...

I'll raise this PR internally to see how we want to handle it.
On the surface it's a very trivial change; however as I eluded to above there's a lot of logic built into the existing readinessProbe to support the nuances of running an Elasticsearch cluster on k8s...

elasticsearch/templates/statefulset.yaml
```diff
@@ -191,8 +191,16 @@ spec:
 {{ toYaml .Values.securityContext | indent 10 }}
         image: "{{ .Values.image }}:{{ .Values.imageTag }}"
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
+      {{- if .Values.readinessProbe.customProbe }}
         readinessProbe:
-{{ toYaml .Values.readinessProbe | indent 10 }}
```
Crazybus: Moving this line to below the currently hardcoded `exec` part will allow all of the keys to be overridden without changing anything else. This change would make the entire readinessProbe configurable without needing to add new logic or values.

With the way that Helm/YAML works it will just take the last key that is specified. So it will allow a custom exec command or the ability to disable the exec command itself by setting it to a null value. I wouldn't recommend it but it still makes sense for this to be possible to override.

mschmidt291: Hi,
yeah I get the point, but the reason for this commit is to specify a httpGet or tcpProbe instead of a exec probe.

Crazybus: Not recommend, but here is how you would replace the exec with a tcpSocket probe instead. To remove a key from the configuration you can set it to a `null` value (https://v2.helm.sh/docs/chart_template_guide/#deleting-a-default-key)

```
readinessProbe:
  exec: null
  tcpSocket:
    port: 9200
  initialDelaySeconds: 5
  periodSeconds: 10
```

I have tested and confirmed that this is working as expected. The only change made is moving this line to below the currently defined exec probe.

mschmidt291: Ah now I understand what you mean. But in my opinion this feels a bit dirty. I like my approach a bit more - it's just cleaner. And since you have to modify the chart in both cases, I would do it my way. 
But as fatmcgav mentioned, this needs to be discussed anyway by elastic.

Crazybus: > But in my opinion this feels a bit dirty. I like my approach a bit more - it's just cleaner. And since you have to modify the chart in both cases, I would do it my way.

Using the documented way to remove a default key for Helm doesn't feel dirty to me. Adding extra templating logic (which will require adding new testing) and duplicating one of the values for this chart feels like the dirtier option. 

> But as fatmcgav mentioned, this needs to be discussed anyway by elastic.

"Discussed by Elastic" is this case roughly translated to "ping @Crazybus to take a look" :)

mschmidt291: okay, i reverted and moved the readinessProbe value to the bottom.

Crazybus:
jenkins test this please

PR #493, opened 2020-02-22
Fix Mismatch Between Service Selector and Pod Labels when using Helm Aliases in Kibana

jamoflaw:

When using aliases to deploy Kibana twice in `requirements.yaml` ie:

```
dependencies:
  - name: kibana
    repository: "@elastic"
    version: 7.6.0
  - name: kibana
    repository: "@elastic"
    version: 7.6.0
    alias: kibana-2
```

Both instances of Kibana will deploy successfully, however the aliased version will be inaccessible as the selector on the service is configured to use the `{{ .Chart.name }}` value which in the case of the aliased chart is different.

This PR allows Kibana to be deployed with an alias in the requirements.yaml

I have added a test which proves this is a noop when deploying in its default state (ie without alias).


jamoflaw:
Just to add I've been using this branch to test and push a version of Kibana which uses this alias and it works well

Am I missing anything from the PR? CLA is signed

jmlrt:
jenkins test this please

jamoflaw:
I've updated the tests fixing the lint issues - running the lint-python target now returns no issues

jmlrt:
jenkins test this please

jamoflaw:
Is the fail a bug in Jenkins? I can't quite see what has failed there, looks as though GKE is complaining about an older version of K8s?

jmlrt:
It seems that GKE has dropped support of 1.13 so we need to drop tests on this version.
You'll need to rebase on master when #533 will be merged.

jamoflaw:
OK pulled in the master branch :+1:

jmlrt:
jenkins test this please

PR #504, opened 2020-03-01
An elasticsearch example for microk8s with security

ncurran:


ncurran:
I've signed the document but I may need to wait a bit longer.

PR #524, opened 2020-03-18
Version bumping script enhancements

Conky5:
Add the ability to switch the namespace and override versions to bump
with environment variables.

~- [ ] Chart version *not* bumped (the versions are all bumped and released at the same time)~
~- [ ] README.md updated with any new values or changes~
~- [ ] Updated template tests in `${CHART}/tests/*.py`~
~- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`~

### DIFF
--- a/helpers/bumper.py
+++ b/helpers/bumper.py
@@ -1,4 +1,16 @@
 #!/usr/bin/env python2
+
+#
+# Usage:
+#   ./bumper.py
+#
+# Configurable environment variables:
+# - BUMPER_VERSION_6 overrides the 6.x.x version.
+# - BUMPER_VERSION_7 overrides the 7.x.x version.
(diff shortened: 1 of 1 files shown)


helpers/bumper.py
```diff
@@ -7,13 +19,13 @@
...
-chart_version = "7.6.1"
-
 versions = {
-    6: "6.8.7",
-    7: "7.6.1",
+    6: os.environ.get("BUMPER_VERSION_6", "6.8.7"),
+    7: os.environ.get("BUMPER_VERSION_7", "7.6.1"),
 }
 
+chart_version = versions[7]
```
jmlrt: Note that we may have to release charts for 6.x also after GA release

PR #542, opened 2020-03-25
Disable service links to prevent very long startup times

floretan:
The Elasticsearch docker image uses environment variables to set Elasticsearch configuration options. Kubernetes automatically adds environment variables for all services in the current namespace. 

When trying to debug why Elasticsearch was taking a particularly long time (over 10 minutes) to boot in one of our projects, we noticed that this behavior was dependent on the namespace, even though no quotas or other limitations were in place. The slow start was also accompanied by a high cpu usage for a few minutes, even though Elasticsearch had no indices at that point. We recently had another issue where the massive number of environment variables was causing an issue, and when we disabled the service links with `enableServiceLinks: false`, the startup behavior went back to normal.

The service links are there to facilitate connecting to other services, but Elasticsearch doesn't usually connect to anything else, so there should be no side effects (when a node connects to another node in the cluster, it uses explicitly declared services, not the automatically generated environment variables). 

I'm not sure if this small static configuration changes needs tests, but I'm happy to add them if needed.


floretan:
This is becoming a blocker for us, it would be very nice to have it merged soon. What would be needed to move forward?

jmlrt:
Hi @floretan, thanks for your PR.
Disable service links make sense for Elasticsearch chart.

`enableServiceLinks` was introduced in K8S 1.13. [Elasticsearch chart README](https://github.com/elastic/helm-charts/blob/master/elasticsearch/README.md#requirements) indicates that it requires K8S >= 1.8.

This PR should update the README requirements and we need to discuss it internally.

jmlrt:
Hi @floretan,

After some internal discussion, we would prefer not breaking the current behavior or compatibility with older K8S versions with Elasticsearch chart default values.

Can you create an `enableServiceLinks` value which would be set to true by default and could be overrided to false?

elasticsearch/templates/statefulset.yaml
```diff
@@ -22,6 +22,7 @@ spec:
   podManagementPolicy: {{ .Values.podManagementPolicy }}
   updateStrategy:
     type: {{ .Values.updateStrategy }}
+  enableServiceLinks: false
```
jmlrt, line 25:
```suggestion
  enableServiceLinks: {{ .Values.enableServiceLinks }}
```

with `enableServiceLinks` set to true by default in `values.yaml`

floretan:
Thank you for taking a closer look at this issue. I hadn't considered the conflict with older kubernetes versions, and I'm glad such attention to details is put into these reviews.

One issue is that `enableServiceLinks: true` would not be a valid syntax for kubernetes < 1.13, even if this is the default behavior. So far I found three options:

- I had a look at using `semverCompare` to check that we have kubernetes 1.13 or greater, but unfortunately I haven't found a way to access the built-in object `.Capabilities.KubeVersion` or `.Capabilities.KubeVersion.Version` as a string that works with both helm 2 and helm 3. Some type checking might work, but the logic wouldn't be very pretty.
- Another solution would be to make `.Values.enableServiceLinks` a configuration option as suggested, but only output the `enableServiceLinks` field if the value is `false`. The default value of `true` in the chart would also match with the default kubernetes behaviour when the field is not specified. This solution would also not be self-explanatory, and would require some comments to explain why things are done this way. 
- A third option would be to simply ignore the fact that `enableServiceLinks: true` would not be valid syntax for kubernetes < 1.13. The schema validation was added in helm 3, and the probability that someone is using helm 3 with a kubernetes version as old as 1.12 is pretty small. 

Which option would you prefer?

jmlrt:
> One issue is that `enableServiceLinks: true` would not be a valid syntax for kubernetes < 1.13, even if this is the default behavior. So far I found three options:

Yep, sorry I didn't think about that, when I reviewed it this morning.

> * I had a look at using `semverCompare` to check that we have kubernetes 1.13 or greater, but unfortunately I haven't found a way to access the built-in object `.Capabilities.KubeVersion` or `.Capabilities.KubeVersion.Version` as a string that works with both helm 2 and helm 3. Some type checking might work, but the logic wouldn't be very pretty.

We are using `semverCompare` with `.Capabilities.KubeVersion.GitVersion` to check K8S version:

https://github.com/elastic/helm-charts/blob/7624ed37a2e5438d15f243887e151b7aba7dcfd0/elasticsearch/templates/_helpers.tpl#L71-L72

This is working fine with Helm 2 but we didn't really tested it with Helm 3 yet. I just saw that it was no more in [Helm 3 doc](https://helm.sh/docs/chart_template_guide/builtin_objects/) and is flagged as deprecated in [Helm 3 code](https://github.com/helm/helm/blob/a28d695c43ad5b889923c5cf0e7694503bdb47df/pkg/chartutil/capabilities.go#L58-L61). 

Note that we officially still don't support Helm 3 so staying with `.Capabilities.KubeVersion.GitVersion` in this PR is fine anyway and if there is an issue with Helm 3 we'll fix it in the same time that we'll fix all the other `.Capabilities.KubeVersion.GitVersion` references.

> * Another solution would be to make .Values.enableServiceLinks a configuration option as suggested, but only output the enableServiceLinks field if the value is false. The default value of true in the chart would also match with the default kubernetes behaviour when the field is not specified. This solution would also not be self-explanatory, and would require some comments to explain why things are done this way.

I'm also totally fine with using an `enableServiceLinks` value set to `true`by default and having a conditional block:

```
{{- if not .Values.enableServiceLinks }}
enableServiceLinks: false
{{- end }}
```

This wouldn't change the current behavior and we would only need to mention that setting this value false is only with K8S >= 1.13 in the [README config description](https://github.com/elastic/helm-charts/blob/master/elasticsearch/README.md#configuration) which could avoid adding using `.Capabilities.KubeVersion.GitVersion`.

floretan:
After taking a closer look at the usage in other charts, I found that `.Capabilities.KubeVersion.GitVersion` is still the only version that I could find in use despite being deprecated in Helm 3, so I went ahead and updated the PR to include a check using this variable.

This gives us the performance improvement out of the box (as mentioned in the original description, there are no side effects), and prevents adding an unnecessary configuration option.

elasticsearch/templates/statefulset.yaml
```diff
@@ -143,6 +143,9 @@ spec:
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
 {{ toYaml .Values.imagePullSecrets | indent 8 }}
+      {{- end }}
+      {{- if semverCompare ">1.13" .Capabilities.KubeVersion.GitVersion }}
+      enableServiceLinks: false
```
jmlrt, line 148: Note that we still want it to be configurable and enabled by default in `values.yaml`in addition of the K8S version check:

```suggestion
      enableServiceLinks: {{ .Values.enableServiceLinks }}
```

floretan: Ok, I made the value configurable and updated both values.yaml and README.md.

jmlrt:
jenkins test this please

floretan:
The build failure seems to be caused by issues when downloading `kubectl`, so it's most likely a temporary network access issue rather than an issue with the changes.

jmlrt:
Yep I'm fine with merging 😄 
Thanks for this PR 👍

jmlrt:

PR #549, opened 2020-03-30
fix: allow redirection on the readinessProbe

kuisathaverat:
## What does this PR do?

It adds the `-L` flag to `curl` to allow redirections.

## Why is it important?

It fixes the readinessProbe for kibana 8.0.0-SNAPSHOT version.

## Related issues
Closes #548

PR #560, opened 2020-04-03
follow redirects for kibana health check

devonwarren:
If you're cluster has authentication enabled the kibana server automatically forwards your traffic to _/login_ via a HTTP 302 redirect which the health check thinks is an error unless you specifically tell it to use the /login path. By using the curl _-L_ flag it will follow the redirect and return the correct code saving the user some debugging


jmlrt:
Hi @devonwarren, this was already fixed by https://github.com/elastic/helm-charts/pull/549/files

devonwarren:
Thanks for the follow-up, will close this

PR #575, opened 2020-04-13
feat: adding liveness probe for kibana

keshavprasadms:


jmlrt:
Hi @keshavprasadms, thank you for this PR, Kibana liveness probe wasn't defined for purpose.
We need to study if we can use the same probe as for readiness.

jmlrt:
cla/check

nikitakazantsev12:
> Hi @keshavprasadms, thank you for this PR, Kibana liveness probe wasn't defined for purpose.
> We need to study if we can use the same probe as for readiness.

Our practical experience shows, that we can use exactly the same probe as a liveness check (now we're achieving that patching existing Kibana-deployment of the Helm-chart).

ecureuil:
Hi, unfortunalately this PR was automatically closed but the feature is still missing. @jmlrt do you want I recreate it, please?

PR #578, opened 2020-04-14
[ci] Fix triggering upstream cleanup job

mgreau:
For doing the integration tests some GKE clusters are created. There is
a dedicated job to clean up these clusters at the end of the tests.
This commit fixes the Jenkins configuration to make sure this upstream
job (which does not need parameter) is triggered.


mgreau:
Failure not related to this update, merging

PR #581, opened 2020-04-15
[script] Update bumper script to handle SNAPSHOT versions

mgreau:
This commit updates the `bumper.py` script to handle the
workflows for updating the versions to test staging artifacts
and before a release, for example:

- update 7.7.0-SNAPSHOT to 7.7.0 before tagging a release, by running
  `env BUMPER_VERSION_7="7.7.0" ./helpers/bumper.py`
- update 7.7.0-SNAPHSHOT to 7.7.0-abcdefgh for testing staging artifacts
by running:
  `env BUMPER_VERSION_7="7.7.0-abcdefgh" BUMPER_USE_STAGING_IMAGES="true" ./helpers/bumper.py`

For now, this script does not handle the bump from 7.7.0 to 7.7.1-SNAPSHOT
for the next development iteration after a release. In the meantime,
this update should be done manually.

### DIFF
--- a/helpers/bumper.py
+++ b/helpers/bumper.py
@@ -28,7 +28,6 @@ chart_version = versions[7]
 file_patterns = [
     "*/examples/*/*.y*ml",
     "helpers/examples.mk",
-    "*/README.md",
     "*/values.y*ml",
     "*/Chart.y*ml",
 ]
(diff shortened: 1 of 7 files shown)


helpers/bumper.py
```diff
@@ -28,7 +28,6 @@
 file_patterns = [
     "*/examples/*/*.y*ml",
     "helpers/examples.mk",
-    "*/README.md",
```
jmlrt, line 31: Why are the README excluded?

We'll need to bump the versions in README to 7.7.0 before the release.

mgreau: Good catch I pushed c54c8c5

PR #586, opened 2020-04-17
[elasticsearch] update readiness probe endpoint

jmlrt:
This PR updates readiness probe endpoint to check only `/` endpoint instead of `/_cluster/health?timeout=0s` when Elasticsearch is already running.
This revert to initial config which was changed in https://github.com/elastic/helm-charts/pull/380 with the exception that 503 HTTP code is accepted for 6.x (see https://github.com/elastic/elasticsearch/issues/8902 for more details about why 503 is OK on Elasticsearch 6.x).

Fix #533

- [ ] ~README.md updated with any new values or changes~
- [ ] ~Updated template tests in `${CHART}/tests/*.py`~
- [ ] ~Updated integration tests in `${CHART}/examples/*/test/goss.yaml`~

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -221,7 +221,6 @@ spec:
                     exit 0
                   elif [[ ${HTTP_CODE} == "503" && "{{ include "elasticsearch.esMajorVersion" . }}" == "6" ]]; then
                     exit 0
-                    exit 0
                   else
                     echo "curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}/ failed with HTTP code ${HTTP_CODE}"
                     exit 1
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -213,7 +213,7 @@ spec:
                   HTTP_CODE=$(curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}/)
                   RC=$?
                   if [[ ${RC} -ne 0 ]]; then
-                    echo "curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}/ failed with RC ${RC}"
+                    echo "curl -XGET -s -k \${BASIC_AUTH} -o /dev/null -w '%{http_code}' {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}/ failed with RC ${RC}"
                     exit ${RC}
                   fi
                   # ready if HTTP code 200, 503 is tolerable if ES version is 6.x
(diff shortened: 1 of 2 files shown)


elasticsearch/templates/statefulset.yaml
```diff
@@ -202,28 +202,37 @@ spec:
...
+                fi
 
                 if [ -f "${START_FILE}" ]; then
-                    echo 'Elasticsearch is already running, lets check the node is healthy and there are master nodes available'
-                    http "/_cluster/health?timeout=0s"
+                  echo 'Elasticsearch is already running, lets check the node is healthy'
+                  HTTP_CODE=$(curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}/)
+                  RC=$?
+                  if [[ ${RC} -ne 0 ]]; then
+                    echo "curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}/ failed with RC ${RC}"
```
Crazybus, line 216: :no_entry_sign: Will this print the BASIC_AUTH details to the Kubernetes event log?

elasticsearch/templates/statefulset.yaml
```diff
@@ -202,28 +202,37 @@ spec:
...
+                    exit 1
```
Crazybus, line 217: :bulb: If we know the actual return code it might be nicer to exit with it explictly rather than always a `1`. 
```suggestion
                    exit ${RC}
```

elasticsearch/templates/statefulset.yaml
```diff
@@ -202,28 +202,37 @@ spec:
...
+                  fi
+                  # ready if HTTP code 200, 503 is tolerable if ES version is 6.x
+                  if [[ ${HTTP_CODE} == "200" ]] || [[ ${HTTP_CODE} == "503" && "{{ include "elasticsearch.esMajorVersion" . }}" == "6" ]]; then
```
Crazybus, line 220: :bulb: How would you feel about splitting up this condition into an elif? It would make it a bit easier to read, and will be much easier to remove once 6.x is EOL.

Given how unexpected the 503 being a passing result is I think including an inline comment linking to the explanation might also help to avoid any future confusion.


Something like:
```suggestion
                  if [[ ${HTTP_CODE} == "200" ]]; then
                    exit 0
                  elif [[ ${HTTP_CODE} == "503" && "{{ include "elasticsearch.esMajorVersion" . }}" == "6" ]]; then
                    exit 0
```

elasticsearch/templates/statefulset.yaml
```diff
@@ -202,28 +202,37 @@ spec:
...
+                    exit 0
+                  else
+                    echo "curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' {{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}/ failed with HTTP code ${HTTP_CODE}"
```
Crazybus, line 223: :no_entry_sign: Same question as above about if these credentials are printed to the event logs.

elasticsearch/templates/statefulset.yaml
jmlrt: Oops thanks for catching that!

elasticsearch/templates/statefulset.yaml
jmlrt: Yeah we can do that! Note that with our new wip branching model, we should soon be able to have only code for `7.x` in `7.7` and `7.x` branchs and code specific to `6.x` in `6.8` branch.

elasticsearch/templates/statefulset.yaml
jmlrt: fixed in a06098c

elasticsearch/templates/statefulset.yaml
jmlrt: fixed in a06098c

PR #591, opened 2020-04-21
[logstash] update doc and values.yaml for http.host issues

jmlrt:
* add warning that `http.host: 0.0.0.0` is required with default probes
* add syntax to disable or override default probes


jmlrt:
backported to `7.7` branch in ab26d11

PR #601, opened 2020-04-28
[meta]  bump version on examples readme

jmlrt:
This PR update bumper script to also update versions found in the examples README files.
`7.0.0-alpha1` version is added to blacklist because we shouldn't update it in [Elasticsearch upgrade example README](https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/upgrade/README.md).


jmlrt:
backported to `7.7` branch in 1c63c35

PR #302, opened 2019-09-28
[elasticsearch] Add compatibility for k8s 1.16

code-chris:
- [x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [ ] README.md updated with any new values or changes
- [ ] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

I don't know if it's possible to add template tests for these changes as they are affected by Values...

PR #635, opened 2020-05-20
[meta] add support for k8s 1.16

jmlrt:
This PR add tests on K8S 1.16 using GKE


jmlrt:
Elasticsearch upgrade test is failling on K8S 1.16 because Elasticsearch chart version < 7.4.0 used `apps/v1beta2` API for `StatefulSet` (changed in #302) which was removed in K8S 1.16 (see [CHANGELOG](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals)).

Elasticsearch upgrade test first install a 7.0.0-alpha1 chart, then upgrade it to last version.

Adding tests and support for K8S 1.16 requires that we update the upgrade test to check 7.4.0 to latest version upgrade instead of 7.0.0-alpha1 to latest version upgrade.

Eventually we could change the test only for K8S 1.16:
- still testing 7.0.0-alpha1 -> latest upgrade for K8S 1.14 and 1.15
- testing 7.4.0 -> latest upgrade for K8S 1.16

elasticsearch/examples/upgrade/README.md
```diff
@@ -1,13 +1,17 @@
 # Upgrade
 
-This example deploy a 3 nodes Elasticsearch cluster using [7.0.0-alpha1][] chart
-version, then upgrade it to 7.7.0 version.
+This example deploy a 3 nodes Elasticsearch cluster using an old chart version,
+then upgrade it to 7.7.0 version.
```
elasticdog, line 4:
```suggestion
This example will deploy a 3 node Elasticsearch cluster using an old chart version,
then upgrade it to version 7.7.0.
```

Minor grammatical tweak.

elasticsearch/examples/upgrade/Makefile
```diff
@@ -3,15 +3,12 @@ default: test
 include ../../../helpers/examples.mk
 
 RELEASE := helm-es-upgrade
+KUBE_MINOR_VERSION := $(shell kubectl version -o json | jq -c '.serverVersion.minor' | sed 's/[^0-9]*//g')
```
elasticdog, line 6: I know this is just an example, but should the requirement for `jq` be documented somewhere?

Conky5: This shell logic seems like a prime candidate to pull out into a real script due to not being able to do error handling here. If `jq` does not exist `7.4.0` will be used. These make files are only used in the testing framework but it could be worth calling out as a dependency.

helpers/matrix.yml
```diff
@@ -40,3 +40,4 @@ APM_SERVER_SUITE:
 KUBERNETES_VERSION:
   - '1.14'
   - '1.15'
+  - '1.16'
```
Conky5, line 43: I wish we could define the chart versions for upgrade testing here somehow so it is very clear what the test cases are, something like this
```yaml
 - kubernetes_version: 1.14
   upgrade_from_chart_version: 7.0.0-alpha1
 - kubernetes_version: 1.15
   upgrade_from_chart_version: 7.0.0-alpha1
 - kubernetes_version: 1.16
   upgrade_from_chart_version: 7.4.0
```
But I think we can only fit this logic into the jenkins job inline shell script, or the Makefile inline shell script the way the job axes work, so adding into the Makefile makes sense to me.

elasticsearch/examples/upgrade/Makefile
jmlrt: I replaced `jq` by `grep` to avoid adding some dependency.

@Conky5, agree with error handling, we can have a script like this to get Elasticsearch version:

```shell
KUBE_MINOR_VERSION=$(kubectl version --client -o yaml | grep minor | sed 's/[^0-9]*//g')

if [ "$KUBE_MINOR_VERSION" -lt 16 ]
then
  echo 7.0.0-alpha1
else
  echo 7.4.0
fi
```

then use `VERSION := $(shell sh ./find-version.sh)` into the Makefile.

WDYT?

Conky5: I would recommending pulling the entire `install` target out into a shell script like `./scripts/install.sh` and including these version computations there.

elasticsearch/examples/upgrade/scripts/upgrade.sh
```diff
@@ -0,0 +1,67 @@
...
+      log "Unrecognized argument: '$key'"
+      usage
+    ;;
+  esac
+done
+
+# Elasticsearch chart < 7.4.0 are not compatible with K8S >= 1.16)
+if [[ -z $FROM ]]
+then
+  KUBE_MINOR_VERSION=$(kubectl version --client -o yaml | grep minor | sed 's/[^0-9]*//g')
```
Conky5, line 49: I think we don't want `--client` since that will show the version of `kubectl` that is being used, which isn't necessarily the server version (especially if someone is trying this out locally), getting the server version would be more accurate.

Re-examining how these jobs and makefiles fit together, I wonder if instead we should pass along the `KUBERNETES_VERSION` that is being passed in by the entrypoint from the jenkins job
https://github.com/elastic/helm-charts/blob/4ece77dd1685ad5a787da9070dc6099417810e53/.ci/jobs/elastic%2Bhelm-charts%2Bstaging%2Bintegration-elasticsearch.yml#L54

Where we could pass it along by updating this to `make KUBERNETES_VERSION=$(KUBERNETES_VERSION)`
https://github.com/elastic/helm-charts/blob/4ece77dd1685ad5a787da9070dc6099417810e53/helpers/terraform/Makefile#L91

Then we could pass this script the KUBERNETES_VERSION with an option like `--kubernetes-version` and determine the elasticsearch version based on that, WDYT?

elasticsearch/examples/upgrade/scripts/upgrade.sh
```diff
@@ -0,0 +1,67 @@
...
+set -eo pipefail
+
+usage() {
+  cat <<-EOF
+	USAGE:
+	  $0 [--release <release-name>] [--from <elasticsearch-version>] --to <elasticsearch-version>
+	  $0 --help
+
+	OPTIONS:
+    --release <release-name>
```
Conky5, line 12: Missed a tab here

elasticsearch/examples/upgrade/scripts/upgrade.sh
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
```
Conky5: Perhaps a `u` as well for good measure.

elasticsearch/examples/upgrade/scripts/upgrade.sh
Conky5: After re-thinking about this, I don't like it, I think it is better to figure out the real server version using `kubectl` somehow.

Conky5: And I think `kubectl ... -o json | jq` is the best way to accomplish it in a script despite it becoming a dependency.

elasticsearch/examples/upgrade/scripts/upgrade.sh
```diff
@@ -43,10 +43,17 @@ do
...
+  echo 'jq is required to use this script'
+  echo 'please check https://stedolan.github.io/jq/download/ to install it'
+  exit 1
+fi
+
 # Elasticsearch chart < 7.4.0 are not compatible with K8S >= 1.16)
 if [[ -z $FROM ]]
 then
-  KUBE_MINOR_VERSION=$(kubectl version --client -o yaml | grep minor | sed 's/[^0-9]*//g')
+  KUBE_MINOR_VERSION=$(kubectl version -o json | jq -c '.serverVersion.minor' | sed 's/[^0-9]*//g')
```
Conky5, line 56: For `jq` I think `--raw-output` could make more sense than `-c`, and we should add `--exit-status` so it'll actually break if it doesn't find the property.

jmlrt:
jenkins test this please

jmlrt:
backported to `6.8`, `7.7`and `7.x` branches

PR #637, opened 2020-05-20
[kibana] String/YAML conditions for `.Values.{extraContainers,extraInitContainers}`

### LINKED ISSUE #636: [kibana] `extraInitContainers` and `extraContainers` as YAMLs
**Describe the feature:**

Currentrly, Kibana chart README says that `extraContainers` and `extraInitContainers` parameters must be the `Templatable string of additional containers to be passed to the tpl function`, but it is conflicting with example values in `values.yaml`:
```
extraContainers: []
  # - name: do-something
  #   image: busybox
  #   command: ['do', 'something']

extraInitContainers: []
  # - name: do-something
  #   image: busybox
  #   command: ['do', 'something']
```
and Helm chart deploy does not work with additional containers defined as YAML manifests.
Elasticsearch chart [has conditions](https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/statefulset.yaml#L394-L403) to accept these values and I think that this can be done in a similar way for Kibana chart.

I would like to be able to define YAML manifests for possible Kibana sidecar containers, for example:

```
extraContainers:
- name: "loader"
  image: "${kibana_image}:${kibana_image_tag}"
  command: ['sleep', 'infinity']
  lifecycle:
    postStart:
      exec:
        command:
        - sh
        - -c
        - |
          #!/usr/bin/env bash -e
          KIBANA_URL=http://localhost:5601
          INIT_FILE=/tmp/.kibana_loader_init
          if [ ! -f "${INIT_FILE}" ]; then
            while [[ "$(curl -XGET -s -o /dev/null -w "%{http_code}" "${KIBANA_URL}/app/kibana")" -ne 200 ]]; do
              sleep 1 && echo "Waiting for Kibana..."
            done
            touch ${INIT_FILE}
            curl -XPOST "${KIBANA_URL}/api/saved_objects/_import" -H "kbn-xsrf: true" \
              --form file=@/some-stuff.ndjson || true
          else
            exit 0
          fi
  volumeMounts:
  - mountPath: /some-stuff.ndjson
    name: kibanaconfig
    subPath: some-stuff.ndjson
```

maxkochubey:
...which are just copied from Elasticsearch helmchart.

Resolves #636.

### DIFF
--- a/kibana/templates/deployment.yaml
+++ b/kibana/templates/deployment.yaml
@@ -69,8 +69,15 @@ spec:
 {{ toYaml .Values.imagePullSecrets | indent 8 }}
     {{- end }}
       {{- if .Values.extraInitContainers }}
+      # Currently some extra blocks accept strings
+      # to continue with backwards compatibility this is being kept
+      # whilst also allowing for yaml to be specified too.
       initContainers:
+      {{- if eq "string" (printf "%T" .Values.extraInitContainers) }}
 {{ tpl .Values.extraInitContainers . | indent 6 }}
+      {{- else }}
+{{ toYaml .Values.extraInitContainers | indent 6 }}
+      {{- end }}
       {{- end }}
       containers:
       - name: kibana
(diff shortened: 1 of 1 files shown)


mitchellmaler:
@maxkochubey Can this be added to the filebeat and metricbeat charts as well? 
All have the extraContainers but it is not yml and using tpl. I also wonder if this should be a new variables like extraContainersYaml that way it doesn't break existing.

jmlrt:
jenkins test this please

jmlrt:
cla/check

jmlrt:
backported to `6.8` branch in ef9aaa5 
backported to `7.x` branch in ba03954 
backported to `7.7` branch in 669e799

Issue #631, opened 2020-05-18
[elasticsearch] New readiness probes causing full cluster-restart


ckotzbauer:
**Chart version:** 7.7.0

**Kubernetes version:** 1.17.1

**Kubernetes provider:** On-prem

**Helm Version:** 3.2.0

**`helm get release` output**

<details>
<summary>Output of helm get release</summary>

```
USER VALUES

esConfig:
  elasticsearch.yml: |
    action.auto_create_index: "-hd-*"
esJavaOpts: -Xmx5g -Xms5g
esMajorVersion: 6
image: "private-image-based-on-official-oss:6.4.2"
imagePullPolicy: IfNotPresent
imagePullSecrets:
- name: some-credentials
imageTag: 6.4.2
ingress:
  annotations:
    ingress.kubernetes.io/ssl-redirect: "true"
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/auth-realm: Authentication Required
    nginx.ingress.kubernetes.io/auth-secret: dev-elasticsearch-ingress-auth
    nginx.ingress.kubernetes.io/auth-type: basic
    nginx.ingress.kubernetes.io/proxy-body-size: 60m
  enabled: true
  hosts:
  - ""
  path: /
  tls:
  - hosts:
    - ""
lifecycle:
  postStart:
    exec:
      command:
      - bash
      - -c
      - |
        #!/bin/bash
        cd /usr/share/elasticsearch/plugins/xxx
        /opt/jdk-10.0.2/bin/jar -cf config.jar config.cfg
        chmod 777 config.jar
persistence:
  enabled: true
podSecurityPolicy:
  create: false
rbac:
  create: true
resources:
  limits:
    cpu: 2000m
    memory: 8Gi
  requests:
    cpu: 200m
    memory: 8Gi
sysctlInitContainer:
  enabled: false
volumeClaimTemplate:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 100Gi
  storageClassName: rook-ceph-cephfs
```

</details>

**Describe the bug:**
I updated the chart to the newest version 7.7.0 and expected that the three-elastic nodes are updated one after another, waiting until the cluster is green again. (The most recent restarted pod was not ready, until the cluster was green again in the past). Now, the pod became ready after a few minutes and kubernetes moved on too quickly, so the cluster was red and not down.

**Steps to reproduce:**

1. Update the chart-installation to 7.7.0

**Expected behavior:**
[...]

Issue #625, opened 2020-05-17
Elasticsearch fails readiness probe due to escaping issue


denysvitali:
The readiness probe for the elasticsearch StatefulSet is implemented without proper escaping:

In my case:
```yaml
apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
  name: elastic
  namespace: monitoring
spec:
  chart: elasticsearch
  repo: https://helm.elastic.co
  targetNamespace: monitoring
  valuesContent: |
    antiAffinity: soft
    roles:
      master: "true"
      ingest: "true"
      data: "true"
    protocol: http
    esConfig:
      elasticsearch.yml: |
        xpack.security.enabled: true
        xpack.security.transport.ssl.enabled: false
        xpack.security.http.ssl.enabled: false
    extraEnvs:
      - name: ELASTIC_PASSWORD
        valueFrom:
          secretKeyRef:
            name: elastic-credentials
            key: password
      - name: ELASTIC_USERNAME
        valueFrom:
          secretKeyRef:
            name: elastic-credentials
            key: username
```

```bash
$ k get secrets elastic-credentials -o jsonpath={.data.password} | base64 -d
S(fMW|(iS6Z"k"@nRK(:
```

(don't worry, the password is a dummy one)

This is the line where I think the error is generated from:
https://github.com/elastic/helm-charts/blob/e0cf8c1137a1c2f39555c948063dbba78db44ef5/elasticsearch/templates/statefulset.yaml#L221

And this is the event log:
```
curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' http://127.0.0.1:9200/ failed with RC 7
8m34s       Warning   Unhealthy                         pod/elasticsearch-master-2                                          Readiness probe failed: Elasticsearch is already running, lets check the node is healthy
curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' http://127.0.0.1:9200/ failed with RC 7
8m33s       Warning   Unhealthy                         pod/elasticsearch-master-0                                          Readiness probe failed: Elasticsearch is already running, lets check the node is healthy
curl -XGET -s -k ${BASIC_AUTH} -o /dev/null -w '%{http_code}' http://127.0.0.1:9200/ failed with RC 7
4m33s       Warning   Unhealthy                         pod/elasticsearch-master-0                                          Readiness probe failed: Elasticsearch is already running, lets check the node is healthy
[...]

PR #626, opened 2020-05-17
[elasticsearch] fix escaping issue in readinessprobe (#625)

denysvitali:
- [X] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [X] README.md updated with any new values or changes
- [ ] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

PR #638, opened 2020-05-20
[elasticsearch] Fix issue with `readinessProbe` causing outages

fatmcgav:
The `readinessProbe` was updated in #586, however this appears to have
introduced a bug which can cause the pod to report as ready before it
actually is, due to a lack of quoting around the URL query params,
combined with the use of '&'.

This commit reworks the `readinessProbe` to correctly quote things.

Reintroduced the `http` function, and added support for passing `args`
in addition to the `path`.
Also borrowed the `set --` pattern from the Kibana chart.

Fixes: #631
Fixes: #625
Related PR: #626

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -4,6 +4,7 @@
 
 
 - [7.7.0 - 2020/05/13](#770---20200513)
+  - [Known Issues](#known-issues)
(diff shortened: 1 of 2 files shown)


fatmcgav:
I've added a `Known issues` entry against the `7.7.0` changelog entry in 4e9c897

PR #494, opened 2020-02-22
added psp policy to filebeat

D1abloRUS:
- [x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [x] README.md updated with any new values or changes
- [x] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

PR #640, opened 2020-05-22
[filebeat] Add pod security policy to template

krichter722:
This is necessary in order to be able to make the chart work on a cluster with pod security policy enabled.


krichter722:
Seems to be already done in #494

PR #657, opened 2020-06-11
[helm] bump helm version to 2.16.8

jmlrt:
Let's bump helm to 2.16.8 which fix a security issue (see [Release Note](https://github.com/helm/helm/releases/tag/v2.16.8))


jmlrt:

PR #659, opened 2020-06-11
[filebeat] introduce dnsConfig values for the containers

orong-pp:
Signed-off-by: Oron Gola <oron.gola@perception-point.io>

### DIFF
--- a/filebeat/values.yaml
+++ b/filebeat/values.yaml
@@ -52,6 +52,9 @@ envFrom: []
 hostPathRoot: /var/lib
 hostNetworking: false
 dnsConfig: {}
+# options:
+#   - name: ndots
+#     value: "2"
 image: "docker.elastic.co/beats/filebeat"
 imageTag: "7.7.1"
 imagePullPolicy: "IfNotPresent"
--- a/filebeat/values.yaml
+++ b/filebeat/values.yaml
@@ -51,6 +51,7 @@ envFrom: []
 # Root directory where Filebeat will write data to in order to persist registry data across pod restarts (file position and other metadata).
 hostPathRoot: /var/lib
 hostNetworking: false
+dnsConfig: {}
 image: "docker.elastic.co/beats/filebeat"
 imageTag: "7.7.1"
 imagePullPolicy: "IfNotPresent"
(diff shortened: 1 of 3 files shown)


orong-pp:
cla/check

orong-pp:
hi
@jmlrt @Crazybus  please review

orong-pp:
### some details:
**the issue:**
when kubernetes runs a dns query in one of its pods, it first tries to look for the record in its internal namespaces. it can take 4-5 queries that search different options before we actually get to the final answer
in case we query an external website like spot.io we get 4-5 times more queries than we actually need and this puts unwanted load on coredns pods.

the details for this are all explained here
https://dev.to/mrkaran/dns-lookups-in-kubernetes-5cm1

**solution:**
so to deal with this issue one can define the ndots dns configuration option
the default is 5 and it means that only urls with 5 or more dots (.) in it will be queried directly from external dns and not go through 4 redundant dns queries

to set the ndots config option lower I added this pull request and made it possible to control this value and set it lower, so that all queries are direct and no internal search is made

**dnsConfig documentation:**
this is how dnsConfig should be configured according to kubernetes docs
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

orong-pp:
I have now signed the CLA

jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

filebeat/values.yaml
```diff
@@ -51,6 +51,7 @@ envFrom: []
 # Root directory where Filebeat will write data to in order to persist registry data across pod restarts (file position and other metadata).
 hostPathRoot: /var/lib
 hostNetworking: false
+dnsConfig: {}
```
jmlrt, line 54: adding an example here would be great
```suggestion
dnsConfig: {}
# options:
#   - name: ndots
#     value: "2"
```

jmlrt (approved):
LGTM⛴, just a suggestion about adding a commented example in `values.yaml`.

orong-pp:
jenkins test this please

orong-pp:
@jmlrt @Crazybus please review

jmlrt (approved):
LGTM⛴

jmlrt:
jenkins test this please

jmlrt:

PR #665, opened 2020-06-15
Elasticsearch: set PVC labels through setting all StatefulSet labels to its volumeClaimTemplate

desaintmartin:
Elasticsearch: set PVC labels through setting all StatefulSet labels to its volumeClaimTemplate.

Allows to track created PVCs (example: delete all PVCs from this recently deleted Release).

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -26,6 +26,16 @@ spec:
   volumeClaimTemplates:
   - metadata:
       name: {{ template "elasticsearch.uname" . }}
+    {{- if .Values.persistence.labels.enabled }}
+      labels:
+        heritage: {{ .Release.Service | quote }}
+        release: {{ .Release.Name | quote }}
+        chart: "{{ .Chart.Name }}"
+        app: "{{ template "elasticsearch.uname" . }}"
+        {{- range $key, $value := .Values.labels }}
+        {{ $key }}: {{ $value | quote }}
+        {{- end }}
+    {{- end }}
     {{- with .Values.persistence.annotations  }}
       annotations:
 {{ toYaml . | indent 8 }}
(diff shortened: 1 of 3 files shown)


jmlrt:
jenkins test this please

desaintmartin:
Sure, here you are. Would you need tests for this?

jmlrt:
jenkins test this please

elasticsearch/templates/statefulset.yaml
```diff
@@ -26,6 +26,16 @@ spec:
   volumeClaimTemplates:
   - metadata:
       name: {{ template "elasticsearch.uname" . }}
+    {{- if .Values.labels.enabled }}
```
jmlrt, line 29:
```suggestion
    {{- if .Values.persistence.labels.enabled }}
```

jmlrt:
> Sure, here you are. Would you need tests for this?

Adding test would be great. Ideally we should have an `assert` in [`test_default`](https://github.com/elastic/helm-charts/blob/master/elasticsearch/tests/elasticsearch_test.py#L13) to ensure that there is no pvc labels with default values, and another test like [`test_adding_storageclass_annotation_to_volumeclaimtemplate`](https://github.com/elastic/helm-charts/blob/master/elasticsearch/tests/elasticsearch_test.py#L453) to ensure that labels are added when they are defined.

desaintmartin:
Updated!

jmlrt:
jenkins test this please

elasticsearch/tests/elasticsearch_test.py
```diff
@@ -84,8 +84,10 @@ def test_defaults():
     assert c["volumeMounts"][0]["mountPath"] == "/usr/share/elasticsearch/data"
     assert c["volumeMounts"][0]["name"] == uname
 
+    # volumeClaimTemplates
     v = r["statefulset"][uname]["spec"]["volumeClaimTemplates"][0]
     assert v["metadata"]["name"] == uname
+    assert v["metadata"]["labels"] == {}
```
jmlrt, line 90: `spec.volumeClaimTemplates.metadata.labels` won't exist in the generated template with default values (if )`persistence.labels.enabled=false`).

```suggestion
    assert "labels" not in v["metadata"]
```

desaintmartin:
Done, signed, rebased.
When running tests locally, it seemed to pass, but I didn't pay attention too much, maybe I did something wrong.

jmlrt:
jenkins test this please

jmlrt:

PR #669, opened 2020-06-17
[elasticsearch] update kind example for version >= 0.7.0

jmlrt:
This PR update the [kubernetes-kind](https://kind.sigs.k8s.io/) example to use new `standard` `StorageClass` instead of the `local-path` `StorageClass` which was added as a workaround for kind < 0.7.0 in #434.


jmlrt:

PR #679, opened 2020-06-19
Update Readme

ArthurFritz:
Update name in the install helm command


jmlrt:
Hi @ArthurFritz, thanks for fixing it 👍

PR #696, opened 2020-06-26
[logstash] fix version in security test

jmlrt:
This commit fixes the expected version in security goss test.


jmlrt:
jenkins test this please

Issue #663, opened 2020-06-12
Logstash don't start with 7.9.0-SNAPSHOT and default values 


jmlrt:
Following `7.x` branch bump to `7.9.0-SNAPSHOT` artifacts in #661, Logstash chart doesn't start anymore with the default values:

```
[2020-06-12T11:46:07,182][ERROR][logstash.monitoringextension.pipelineregisterhook] Failed to set up the metrics pipeline {:message=>"With the default xpack.monitoring.elasticsearch.username, xpack.monitoring.elasticsearch.password must be set", :backtrace=>["/usr/share/logstash/x-pack/lib/helpers/elasticsearch_options.rb:207:in `validate_authentication!'", "/usr/share/logstash/x-pack/lib/helpers/elasticsearch_options.rb:50:in `es_options_from_settings'", "/usr/share/logstash/x-pack/lib/helpers/elasticsearch_options.rb:41:in `es_options_from_settings_or_modules'", "/usr/share/logstash/x-pack/lib/monitoring/monitoring.rb:201:in `generate_pipeline_config'", "/usr/share/logstash/x-pack/lib/monitoring/monitoring.rb:180:in `setup_metrics_pipeline'", "/usr/share/logstash/x-pack/lib/monitoring/monitoring.rb:142:in `after_agent'", "org/logstash/execution/EventDispatcherExt.java:91:in `execute'", "/usr/share/logstash/logstash-core/lib/logstash/runner.rb:388:in `execute'", "/usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/clamp-0.6.5/lib/clamp/command.rb:67:in `run'", "/usr/share/logstash/logstash-core/lib/logstash/runner.rb:268:in `run'", "/usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/clamp-0.6.5/lib/clamp/command.rb:132:in `run'", "/usr/share/logstash/lib/bootstrap/environment.rb:88:in `<main>'"]}
[...]

PR #697, opened 2020-06-26
[logstash] fix version in security test

jmlrt:
This commit fixes the expected version in security goss test.


jmlrt:
Logstash default issues are related to https://github.com/elastic/helm-charts/issues/663.
Logstash security tests are OK with this fix.
This is ready to merge.

PR #708, opened 2020-07-03
[apm-server] Add to test matrix

fatmcgav:
Spotted that the `apm-server` chart isn't being tested by PR suite.

So add it to the `CHART` matrix.


jmlrt:
Logstash errors are related to #663. This is ready to merge

Issue #702, opened 2020-07-01
Missing option to disable the daemonset in metricbeat


erihanse:
**Chart version:**
7.8.0
**Kubernetes version:**
1.15.9
**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
on-premise
**Helm Version:**
2.12.2

**`helm get release` output**

<details>
<summary>Output of helm get release</summary>

```
REVISION: 1
RELEASED: Tue Jun 30 20:48:16 2020
CHART: elasticsearch-7.3.2
USER-SUPPLIED VALUES:
clusterHealthCheckParams: wait_for_status=yellow&timeout=10s
minimumMasterNodes: 1
nodeSelector:
  kubernetes.io/hostname: demoworker2test
podSecurityPolicy:
  name: 50-rootfilesystem
rbac:
  create: true
  serviceAccountName: elasticsearch-master
readinessProbe:
  initialDelaySeconds: 30
  timeoutSeconds: 20
replicas: 1
resources:
  limits:
    memory: 2.5Gi
  requests:
    memory: 1Gi
sysctlInitContainer:
  enabled: false

COMPUTED VALUES:
antiAffinity: hard
antiAffinityTopologyKey: kubernetes.io/hostname
clusterHealthCheckParams: wait_for_status=yellow&timeout=10s
clusterName: elasticsearch
esConfig: {}
esJavaOpts: -Xmx1g -Xms1g
esMajorVersion: ""
extraEnvs: []
extraInitContainers: ""
extraVolumeMounts: ""
extraVolumes: ""
fsGroup: ""
fullnameOverride: ""
httpPort: 9200
image: docker.elastic.co/elasticsearch/elasticsearch
imagePullPolicy: IfNotPresent
imagePullSecrets: []
imageTag: 7.3.2
ingress:
  annotations: {}
  enabled: false
  hosts:
  - chart-example.local
  path: /
  tls: []
initResources: {}
keystore: []
labels: {}
lifecycle: {}
masterService: ""
masterTerminationFix: false
maxUnavailable: 1
minimumMasterNodes: 1
nameOverride: ""
networkHost: 0.0.0.0
nodeAffinity: {}
nodeGroup: master
nodeSelector:
  kubernetes.io/hostname: demoworker2test
persistence:
  annotations: {}
  enabled: true
podAnnotations: {}
podManagementPolicy: Parallel
podSecurityContext:
  fsGroup: 1000
  runAsUser: 1000
[...]

PR #715, opened 2020-07-07
[testing] Ensure that the 'namespace' is specified in Helm command

fatmcgav:
If the `--namespace` flag isn't provided to Helm, then it will use whatever the local configured namespace is, which could be different.

This can lead to tests failing that expect the `.Release.Namespace` to resolve to `default`


fatmcgav:
This came up from https://github.com/elastic/helm-charts/issues/702#issuecomment-654910011

PR #728, opened 2020-07-15
[meta] use same staging job for all branches

jmlrt:
We currently have staging jobs for every branch (6.8, 7.8, 7.x & master). However, we should only test staging from the minor release branches (ie: 6.8 & 7.8). In addition, staging jobs are only triggered manually after a BC, so staging job for a specific branch should rarely have more than 10 builds. This commit will rationalize and help having a cleaner Helm charts Jenkins view by using the same staging jobs for every branches. 

Note that staging jobs templates should be removed from the other branches.

### DIFF
--- a/.ci/jobs/elastic+helm-charts+staging+cluster-cleanup.yml
+++ b/.ci/jobs/elastic+helm-charts+staging+cluster-cleanup.yml
@@ -6,9 +6,14 @@
     parameters:
       - string:
           name: BUILD_ID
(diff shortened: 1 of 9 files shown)


jmlrt:
Current list of staging jobs in Jenkins [Helm charts view](https://devops-ci.elastic.co/view/Helm%20Charts/):
Screenshot 2020-07-15 at 15 24 49

PR #729, opened 2020-07-16
Include pre-releases in the semver range.

floretan:

The new parameter `enableServiceLinks` added in https://github.com/elastic/helm-charts/pull/542 is only rendered with kubernetes >1.13. However the notation for the parameter to `semverCompare` requires a `-0` suffix to also take into account versions like `v1.15.12-gke.6.12`. Because of that, the `enableServiceLinks` parameter currently nothing on GKE. 

Note that the `helm template` and `helm install` will generate different values for `.Capabilities.KubeVersion.GitVersion`, something I was previously not aware of.

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -154,7 +154,7 @@ spec:
       imagePullSecrets:
 {{ toYaml .Values.imagePullSecrets | indent 8 }}
       {{- end }}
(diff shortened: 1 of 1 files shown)


jmlrt:
jenkins test this please

jmlrt (approved):
LGTM⛴

jmlrt:

PR #736, opened 2020-07-17
[meta] remove branch specific staging jobs

jmlrt:
These jobs are replaced by a single job which can manage every branch in 47a45cf

This PR is following #728 and will be backported in the other release branches (`7.8`, `7.x`).


jmlrt:
backported to `7.8`, `7.9` and `7.x` branches

PR #755, opened 2020-07-27
[meta] update changelog for 7.8.1 and 6.8.11 releases

jmlrt:
This PR update the changelog following #753 & #754

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -3,9 +3,9 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
 
-- [7.8.1 - 2020/XX/XX](#781---2020xxxx)
+- [7.8.1 - 2020/07/28](#781---20200728)
     - [Add headless Service for StatefulSet](#add-headless-service-for-statefulset)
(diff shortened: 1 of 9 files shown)


BREAKING_CHANGES.md
```diff
@@ -29,6 +32,26 @@
...
+A headless `Service` has been added to Logstash chart in [#695][].
+
+Headless `Service` is required for `Statefulsets`. While Helm 2 allowed to
+deploy `Statefulset` without a `serviceName`, however Helm 3 enforce this
+requirement and fails if `serviceName` is missing.
+
+`Statefulset` don't accept `serviceName` field update during release upgrades.
+Upgrading Logstash chart from a previous version will require using
+`helm upgrade --force`.
+
```
Conky5, line 48: Minor grammar tweaks for these last 2 paragraphs

```
The headless `Service` is required for `Statefulsets`. Helm 2 allowed
deploying a `Statefulset` without a `serviceName`, however Helm 3 enforces this
requirement and fails if `serviceName` is missing.

`Statefulset` does not accept the `serviceName` field update during release upgrades.
Upgrading the Logstash chart from a previous version will require using
`helm upgrade --force`.
```

jmlrt: updated in 2fbfcf0299a794b2bb35db04a38da27a064e429f

PR #759, opened 2020-07-28
[meta] add helm 3 beta support

jmlrt:
This PR add Helm 3 support in preview as the main blockers were fixed in previous PR.

Note that adding automated tests with Helm 3 is still required to add full Helm 3 support.

This PR should be backported to `6.8`, `7.x` and `7.9` branches.


jmlrt:
This PR add Helm 3 support in beta as the main blockers were fixed in previous PR.

Note that adding automated tests with Helm 3 is still required to add full Helm 3 support.

jmlrt:

PR #802, opened 2020-09-03
Adding Podmonitor for filebeat

vinothayyavoo:

### DIFF
--- a/filebeat/templates/podmonitor.yaml
+++ b/filebeat/templates/podmonitor.yaml
@@ -1,0 +1,31 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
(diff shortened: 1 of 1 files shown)


jmlrt:
Hi @vinothayyavoo, thank you for this PR.
`PodMonitor` isn't a Kubernetes standard but is specific to Prometheus.
We may revisit it later, but want to avoid managing integration with third party resources outside of Elastic products and Kubernetes standard resources in Elastic charts currently.

vinothayyavoo:
@jmlrt I understand Pod/Service monitor are relevant to Prometheus. The reason for the PR here is to get everything at one place and in the same standard instead of using charts from multiple locations.

It would be nice to have the support of monitors here.

PR #805, opened 2020-09-03
Filebeat

vinothayyavoo:


jmlrt:
Same as #802. Closing is because we don't intend to add Prometheus integration to Elastic Helm charts.

PR #831, opened 2020-10-01
7.9.3 snapshot

nkammah:
Bump chart version and Docker images to 7.9.3-SNAPSHOT to test 7.9.3 development version

See [New branching model](https://github.com/elastic/helm-charts/blob/master/BREAKING_CHANGES.md#new-branching-model) for more details.


apm-server/README.md
```diff
@@ -40,17 +41,26 @@ See [supported configurations][] for more details.
...
+  - with [Helm 3 (beta)][]: `helm install apm-server --version <version> elastic/apm-server`
+
+### Install development version from a branch
+
+* Clone the git repo: `git clone git@github.com:elastic/helm-charts.git`
 
+* Checkout the branch : `git checkout 7.9`
+
+* Install it:
+  - with Helm 2: `helm install --name apm-server ./helm-charts/apm-server`
```
jmlrt, line 62: it should be specified in this section that it is also using snapshot docker images to avoid users confusion.

jmlrt: Note that this already specified in https://github.com/elastic/helm-charts/pull/831/files#diff-ae72c3cb6c31c49e02d8b7896a43c68fR44:

>This chart is tested with the latest 7.9.3-SNAPSHOT version.

However, that may not be clear enough for users

jmlrt: we may add `--set imageTag=<version>` in the command...

apm-server/README.md
```diff
@@ -40,17 +41,26 @@ See [supported configurations][] for more details.
...
-  - with Helm 2: `helm install --name apm-server --version 7.9.2 elastic/apm-server`
-  - with [Helm 3 (beta)][]: `helm install apm-server --version 7.9.2 elastic/apm-server`
+  - with Helm 2: `helm install --name apm-server --version <version> elastic/apm-server`
+  - with [Helm 3 (beta)][]: `helm install apm-server --version <version> elastic/apm-server`
+
+### Install development version from a branch
+
+* Clone the git repo: `git clone git@github.com:elastic/helm-charts.git`
 
+* Checkout the branch : `git checkout 7.9`
```
jmlrt, line 59: good point, this was missing previously 👍

apm-server/README.md
```diff
@@ -40,17 +41,26 @@ See [supported configurations][] for more details.
...
+
+* Install it:
+  - with Helm 2: `helm install --name apm-server ./helm-charts/apm-server`
+  - with [Helm 3 (beta)][]: `helm install apm-server ./helm-charts/apm-server`
```
jmlrt, line 63: What is the point of removing the `7.9.3-SNAPSHOT` from this section.
Should we expect not removing it from the readme when doing releases?

jmlrt: Oh yes, it seems it was added in the previous pr but not in the doc 👍

jmlrt (changes requested):
Test for Filebeat, Metricbeat, Kibana and Logstash are failing because version shouldn't have the `-SNAPSHOT` suffix in their respective goss yaml files.

jmlrt (approved):
LGTM⛴

Issue #807, opened 2020-09-08
[logstash] Unable to install Logstash with configured nodePort


Gagarmel:
**Chart version:**
7.9.1

**Kubernetes version:**
v1.15.12

**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
Own

**Helm Version:**
3.3.0

**Provided values.yaml**
```
service:
  type: NodePort
  ports:
    - name: beats
      port: 5044
      protocol: TCP
      nodePort: 31545
```

**Describe the bug:**
Due to changes made in #776 it is not possible to install a Logstash instance which has a configured `nodePort`.

**Steps to reproduce:**

1. Use the above mentioned `values.yaml`
2. Install Logstash `helm install logstash elastic/logstash -f values.yaml`

**Expected behavior:**
It should be possible to create a Logstash installation with Service `type=NodePort` and configured `nodePort=xxxxx`. This was possible in previous versions of this chart. 

**Provide logs and/or server output (if relevant):**

*Be careful to obfuscate every secrets (credentials, token, public IP, ...) that could be visible in the output before copy-pasting*

```
Error: Service "logstash-logstash-headless" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'
```

**Any additional context:**

jmlrt:
Hi @Gagarmel,
Thanks for reporting this bug 👍 

Good catch, I didn't test it with `NodePort` when I reviewed #776 😞.

IMO we should be able to remove totally `.Values.service.ports` and use only the beat port for the headless service as the headless service should only be used for K8S statefulset management and a real service should be created for other usages.

https://github.com/elastic/helm-charts/blob/6ddd3e36b066ab7e116b3d2810918dc39223e3c8/logstash/templates/service-headless.yaml#L19-L21

PR #839, opened 2020-10-14
[logstash] use only httpPort in headless service

jmlrt:
This commit fix an issue introduced in 8ed75a9 where helm
install/upgrade is failing if a NodePort service is defined.

Logstash headless service used to manage Statefulset does not anymore include
`.Values.service.ports` and include only Logstash httpPort.

This is done because headless service `ports` doesn't accept `nodePort`
values that can be included in `.Values.service.ports` when adding a
NodePort service.

Fix https://github.com/elastic/helm-charts/issues/807

### DIFF
--- a/logstash/templates/service-headless.yaml
+++ b/logstash/templates/service-headless.yaml
@@ -16,9 +16,5 @@ spec:
   selector:
     app: "{{ template "logstash.fullname" . }}"
   ports:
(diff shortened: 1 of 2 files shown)


Conky5 (approved):
LGTM

jmlrt:

Issue #832, opened 2020-10-02
Deployed Coordinator nodes only not set with the ES helm chart


bamb00:
In the ES helm chart all node roles is set to "false" with replicas '3'.  Querying the cluster nodes...I don't see the '-' assigned to the coordinator node.  The coordinator nodes name is elasticsearch-coordinator-0,1,2.  I've also noticed pvc was created for coordinator nodes only.

From the output below I could not determined if the three nodes (elasticsearch-coordinator-*) is designate as an coordinator only nodes.  I look at the pods log to verify if the nodes were coordinator but no infos returned.  Also, How do I test or run command to verify if is a coordinator node?

```
$ kubectl exec elasticsearch-coordinator-0 -- curl -X GET 'http://localhost:9200/_cat/nodes?pretty'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current  Dload  Upload   Total   Spent    Left  Speed                                 
100   649  100   649    0     0  24571      0 --:--:-- --:--:-- --:--:-- 24961
10.2.1.101 41 25 2 0.00 0.00 0.00  lr - elasticsearch-coordinator-0
10.2.1.72  46 25 3 0.00 0.00 0.00  lr - elasticsearch-coordinator-1
10.2.1.129 54 25 3 0.07 0.12 0.10  lr - elasticsearch-coordinator-2
10.2.0.236 16 46 1 0.00 0.01 0.04  ilr - elasticsearch-ingest-1
10.2.1.44  38 68 2 0.09 0.17 0.13  dlrt - elasticsearch-data-1
10.2.1.23  12 46 1 0.02 0.09 0.09  ilr - elasticsearch-ingest-0
10.2.0.106 32 68 2 0.18 0.07 0.08 dlrt - elasticsearch-data-2
10.2.0.88  40 47 2 0.88 0.47 0.23 lmr * elasticsearch-master-0
10.2.0.174 26 68 2 0.05 0.09 0.10 dlrt - elasticsearch-data-0
10.2.0.20  44 46 2 0.72 0.36 0.28 lmr - elasticsearch-master-1
```

jmlrt:
Hi @bamb00,
We generally recommend that questions such as this are asked on our Discuss forums at discuss.elastic.co.

dcvtruong:
Hi @jmlrt, When I post question related to helm chart? Most of the time no response from the forum.  Since this is a helm related configuration deployment I'm almost positive there would be no response.

[...]

PR #854, opened 2020-10-21
[elasticsearch] add coordinator node to multi test

jmlrt:
This commit add a coordinator node group to elasticsearch multi test.
This also add ml and remote_cluster_client roles to default values.

Related to #832


Conky5 (approved):
LGTM

PR #882, opened 2020-10-30
[metricbeat] use relocated stable repo for kube-state-metrics

jmlrt:
This commit update metricbeat chart dependencies to get
kube-state-metrics subchart from the newly relocated stable repo url
(https://charts.helm.sh/stable).

This is required because the legacy stable repo url will not work
anymore starting November 13th.

More details in https://helm.sh/blog/new-location-stable-incubator-charts/


Conky5 (approved):
LGTM

jmlrt:

PR #885, opened 2020-11-05
[meta] clean deprecated helpers tools and docs

jmlrt:
This commit remove the release script and process documentation
previously used to release Helm charts. They have been replaced
by a new automated process available in Elastic private repos

Note that the helpers/bumper.py script isn't used anymore for release
bumps but is still used for staging tests.


nkammah (approved):
lgtm 🚢

PR #891, opened 2020-11-10
[meta] increase helm timeout

jmlrt:
This commit increase the timeout for helm upgrade commands examples to
avoid timeout error when deploying all example releases in // in helm
test gke cluster.


Conky5 (approved):
LGTM

jmlrt:

Issue #895, opened 2020-11-12
[filebeat] - Support ServiceMonitor for metrics collection


ryan-dyer-sp:
**Describe the feature:**
Add a servicemonitor template to the helm chart so those using prometheus can easily scrape metrics using the filebeat helm chart.

**Describe a specific use case for the feature:**
As a user of the filebeat helm chart I want a way for my prometheus cluster to readily scrape the metrics from filebeat.  As filebeat does not have native openmetrics support, I have added the trustpilot/beats-exporter as an extraContainer via the helm chart.  Now I just need to scrape that endpoint using a prometheus ServiceMonitor.

Will submit PR shortly with changes.

jmlrt:
Hi @vinothayyavoo, thank you for this feature request.
`ServiceMonitor` isn't a Kubernetes standard but is specific to Prometheus.
We may revisit it later, but want to avoid managing integration with third party resources outside of Elastic products and Kubernetes standard resources in Elastic charts currently.

Issue #231, opened 2019-07-22
Prometheus-operator support


desaintmartin:
Hello guys,
Do you plan to support prometheus-operator specific things like ServiceMonitor or PrometheusRule? If so, would you accept PR for this?

Crazybus:
No plans to do this. It also feels kind of weird to manage thirdPartyResources for a specific monitoring tool directly in an unrelated helm chart. It seems like most of the charts doing this are actually hard-coding the monitoring rules into the chart too. Going down this path would mean there would be a future where a whole bunch of thirdPartyResources were being defined and managed. Not only would that be very hard to test and support, it would also get really messy. 

How would you feel about having a more generic value that allows a user to specify whatever templated resources they want? This would allow them to specify thirdPartyResources with templating and define it exactly how they want without adding any extra complexity to the chart.

desaintmartin:
I understand.

With the current state of helm, wouldn't adding thirdPartyResources with templating become a mess to maintain and understand?

Crazybus:
So I didn't mean specifically trying to template thirdPartyResources. More just giving a free form variable for users to dump other templatable resources for it to be managed by the chart itself.

So having a separate file that includes a variable which can contain templating: 
```
{{ tpl .Values.extraYaml . }}
```

The user could then do something like (this is just a copy pasted example that isn't tested)

```
extraYaml: |
  ---
  apiVersion: monitoring.coreos.com/v1
  kind: PrometheusRule
  metadata:
    name: {{ template "fullname" . }}
    labels:
      "app.kubernetes.io/name": '{{ template "name" .}}'
      "helm.sh/chart": "{{ .Chart.Name }}-{{ .Chart.Version }}"
      "app.kubernetes.io/managed-by": "{{ .Release.Service }}"
      "app.kubernetes.io/instance": "{{ .Release.Name }}"
      "app.kubernetes.io/component": "{{ .Values.master.componentName }}"
  spec:
    groups:
  {{ toYaml .Values.prometheus.alertingrules | indent 2 }}
```

So nothing specific about thirdPartyResources or PrometheusRule.

desaintmartin:
After a bit of thoughts, deploying an [elasticsearch-exporter](https://github.com/helm/charts/tree/master/stable/elasticsearch-exporter) does the trick, and having a hat chart works well (with a few hardcodes in the values.yaml). Closing.

PR #896, opened 2020-11-12
895

ryan-dyer-sp:
- [no] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

Did not add any integration tests for this as I'm not sure how goss runs.  If it runs by attempting to actually create a k8s object then this would likely fail unless the cluster in question has the servicemonitor crd.

Fixes Issue #895


ryan-dyer-sp:
CLA has been signed.

jmlrt:
Hi @vinothayyavoo, thank you for this feature request.
`ServiceMonitor` isn't a Kubernetes standard but is specific to Prometheus.
We may revisit it later, but want to avoid managing integration with third party resources outside of Elastic products and Kubernetes standard resources in Elastic charts currently.

ryan-dyer-sp:
@jmlrt I assume you meant me.

What about the headless service part?  Every piece that can be added to support the usecase makes it easier for users to monitor filebeats.

And though I can understand the opinion, its not one that is shared among many helm charts.  The majority of people running kubernetes, also run prometheus and want a way to scrape metrics.  Filebeats doesnt support an openmetrics standard (even though there's been a ticket open for a long time asking for this) and the extraContainers aspect of the helm chart is sufficient for adding this functionality.  Would you support a generic extraManifests value?  Where it could be an array of manifests that get added?  That way we could add whatever additional objects were needed to make filebeat fully integrate into our environments?

jmlrt:
Hi @ryan-dyer-sp,
First I'm sorry for the name mistake.

I'm not sure to get the point about headless service.

Elastic recommands using our Metricbeat chart to get pods metrics and inject them into Elasticsearch, that's another reason if needed why we won't add Prometheus integration into our charts.

Adding some `extraManifest` or `extraYaml` value has been discussed in a similar issue in https://github.com/elastic/helm-charts/issues/231#issuecomment-514235426. It's something that we could merge if you're interested in submitting a PR.

ryan-dyer-sp:
@jmlrt The metrics being available in elasticsearch isnt of any use to us at this time, we want them in prometheus.  

For a servicemonitor to work you need a headless service in order for prometheus to determine the various pod IPs (via the endpoints) to scrape.  The argument was elastic doesnt want to support non-native k8s objects.  A headless service is a native k8s construct.  So would you prefer to keep the headless service creation (and change the field to be .Values.service.enabled)  or do you prefer to keep it all out of the helm chart entirely and just allow users to do it via an extraManifests if they need it?

I will modify the PR to add an extraManifest field as a way to inject additional manifests into the helm chart, thanks!

PR #516, opened 2020-03-13
Helm 3

jmlrt:
This is a draft PR to run all CI tests with Helm 3

- [ ] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [ ] README.md updated with any new values or changes
- [ ] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

Related to #401 
Fix #492

PR #898, opened 2020-11-16
[meta] upgrade test

### LINKED ISSUE #396: Add upgrade test for every chart
Upgrading kube-state-metrics Metricbeat dependency in #352 introduced a breaking change which make helm upgrade failing (#395).

Having automated upgrade tests for all charts like the one we already have for [elasticsearch](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples/upgrade) could help us to detect these breaking changes before merging.

jmlrt:
This PR is adding a new upgrade test for every chart.

This test is deploying the oldest **release** version of charts which can be upgraded to latest version, then upgrading using unreleased charts in the repo.

The goal is to detect when some PR is introducing a breaking change.

- Note 1: An upgrade test is already existing for Elasticsearch but had been removed from master branch because upgrade from 7.x to Elasticsearch 8.0.0-SNAPSHOT used by master branch isn't working. The new test allows overriding the Docker image version used during upgrade so we can still test upgrade on master but use the latest 7.x Docker image instead of 8.0.0-SNAPSHOT.

* Note 2:  Metricbeat chart introduced a breaking change in #516 and can be upgraded from any previous version currently, so Metricbeat test is commented and should be uncommented after 7.10.0 release so we can check if new PR are introducing new breaking changes.

* Note 3: Currently charts upgrade are only tested with default values and so cover a small part of charts codebase, we should update it to test optional resources (ie: enable service or ingress, ...).

Fix #396


jmlrt:
jenkins test this please

jmlrt:
replaced by #907

PR #906, opened 2020-11-17
[elasticsearch] allow specifying extraContainers as a map

### LINKED ISSUE #874: Elasticsearch: Allow specifying extraContainers as a map
**Describe the feature:**
Currently, the Elasticsearch chart supports specifying `extraContainers` as a YAML array (or a templatable string for backward compatibility). This is fine for simple use cases, but not convenient if overriding via multiple values files is desired.

This feature request is to additionally support providing `extraContainers` as a map, which would facilitate overrides through multiple layers of values files.

Consider a simple `my-values.yaml` file:

``` yaml
extraContainers:
- name: do-something
  image: my-image
  command: ['do', 'something']
```

I want to make use of helm's file merging to override the image name in subsequent deploy steps using a separate `my-values.override.yaml` file. Under the current implementation, this will require me to duplicate the entire spec:

``` yaml
extraContainers:
- name: do-something
  image: my-override-image
  command: ['do', 'something']
```

This repetition is not convenient and will cause a lot of maintenance issues. The proposed implementation would be to allow expressing `my-values.yaml` like so:

``` yaml
extraContainers:
  something:
    name: do-something
    image: my-image
    command: ['do', 'something']
```

Which could then be overridden by another `my-values.override.yaml`:

``` yaml
extraContainers:
  something:
    image: my-override-image
```

musabhusaini:
- add helper function to convert arrays, map values, or templatable strings to arrays
- add support for providing elasticsearch extraContainers as a map
- add test for providing extraContainers as a map
- update values.yaml to document providing extraContainers as a map
- update README.md to document providing extraContainers as a map

This PR addresses #874 by allowing one to specify `extraContainers` as a map in addition to the currently supported array and templatable string. This facilitates overriding values through multiple layers of values files as discussed in the above-mentioned issue.

The following YAML snippet demonstrates how to use this format:

``` yaml
extraContainers:
  something:
    name: do-something
    image: my-image
    command: ['do', 'something']
```

Which is equivalent to:

``` yaml
extraContainers:
- name: do-something
  image: my-image
  command: ['do', 'something']
```

Fix #874


elasticsearch/templates/_helpers.tpl
```diff
@@ -85,3 +85,25 @@ Return the appropriate apiVersion for ingress.
 {{- print "networking.k8s.io/v1beta1" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Returns the provided array, map values, or templatable string as an array.
+*/}}
+{{- define "elasticsearch.statefulset.toArray" -}}
```
musabhusaini, line 92: I had initially written this inlined in the `./statefulset.yaml` template, but @jmlrt mentioned that other places could benefit from this as well. I don't want to expand the scope of this PR, but future PRs can use this to make `extraInitContainers` etc. more flexible.

jmlrt:
Hi @musabhusaini,
Thanks for opening this PR.
We'll try to review it soon. Please be patient.

jmlrt:
still need to find time for review

jmlrt:
jenkins test this please

jmlrt:
still valid

PR #907, opened 2020-11-17
[meta] upgrade test

### LINKED ISSUE #396: Add upgrade test for every chart

jmlrt:
This PR is adding a new upgrade test for every chart.

This test is deploying the oldest **release** version of charts which can be upgraded to latest version, then upgrading using unreleased charts in the repo.

The goal is to detect when some PR is introducing a breaking change.

- Note 1: An upgrade test is already existing for Elasticsearch but had been removed from master branch because upgrade from 7.x to Elasticsearch 8.0.0-SNAPSHOT used by master branch isn't working. The new test allows overriding the Docker image version used during upgrade so we can still test upgrade on master but use the latest 7.x Docker image instead of 8.0.0-SNAPSHOT.

* Note 2:  Metricbeat chart introduced a breaking change in #516 and can be upgraded from any previous version currently, so Metricbeat test is commented and should be uncommented after 7.10.0 release so we can check if new PR are introducing new breaking changes.

* Note 3: Currently charts upgrade are only tested with default values and so cover a small part of charts codebase, we should update it to test optional resources (ie: enable service or ingress, ...).

Fix #396


jmlrt:
Failing tests are new transient issues not related to this PR.

helpers/matrix.yml
```diff
@@ -11,28 +11,34 @@ ES_SUITE:
...
 FILEBEAT_SUITE:
   - default
   - oss
   - security
+  - upgrade
 METRICBEAT_SUITE:
   - default
   - oss
   - security
+  - upgrade
```
Conky5, line 29: I see the `FROM` is commented out in the metricbeat upgrade makefile, is it ready to be added without the 7.10.0 helm-chart release being out yet?

elasticsearch/examples/upgrade/Makefile
```diff
@@ -0,0 +1,16 @@
+default: test
+
+include ../../../helpers/examples.mk
+
+CHART := elasticsearch
+RELEASE := helm-es-upgrade
+FROM := 7.4.0	# versions before 7.4.O aren't compatible with Kubernetes >= 1.16.0
+TO := 7.10.0	# upgrade from 7.x to 8.0.0-SNAPSHOT currently doesn't work
```
Conky5, line 8: Are there any issues with the version numbers in these files getting updated by the version bumping automation?

helpers/matrix.yml
jmlrt: Actually the `test` target is also commented in https://github.com/elastic/helm-charts/pull/907/files?file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension#diff-24429fd964f0f618b6c9c57dd7ea1977aa62316fbce76bef1025b24a1da762edR14

So the Metricbeat upgrade will pass without doing anything for now. until we release 7.10.0 then uncomment `FROM` and `test` target.

Conky5: Ah I see, I think it would be good to remove it from this file as well to avoid spinning up a worker/job that doesn't actually do anything and gives the appearance of passing.

elasticsearch/examples/upgrade/Makefile
jmlrt: Good question!
- On master branch we don't do any bumping so we won't have any issues for sure.
- On release branches, we shouldn't need have issue because [community-release-tools](https://github.com/elastic/infra/blob/master/cd/release/community-release-tools) get old version from `Charts.yaml` and only bump this version.
  In addition we shouldn't need to set `TO` on release branches because release branches are using SNAPSHOT images from the same major stack version as the `FROM` version unlike master branch which is using 8.0.0 SNAPSHOT images.

helpers/matrix.yml
jmlrt: done in e7a2df5

jmlrt:
Failing tests are still transient errors.

Conky5 (approved):
LGTM! :ferry:

PR #916, opened 2020-11-19
[meta] remove support for k8s <1.14 & helm <2.17.0

jmlrt:
This commit remove the code support for k8s version < 1.14.
This allows us to remove some helpers which aren't working anymore with
Helm 2.

```
Error: render error in "logstash/templates/statefulset.yaml": template: logstash/templates/_helpers.tpl:26:43: executing "logstash.statefulset.apiVersion" at <.Capabilities.KubeVersion.Version>: can't evaluate field Version in type *version.Info
```

In addition this also remove support for helm version < 2.17.0 which is
the last Helm 2 version.


Conky5 (approved):
LGTM

jmlrt:
jenkins test this please

PR #939, opened 2020-11-25
[meta] allows concurrent jobs for pr

jmlrt:
This commit update PR jobs to allow concurrency. Only the multijob
allowed concurrency but not the downstream jobs


Conky5 (approved):
LGTM :ferris_wheel:

PR #940, opened 2020-11-25
[meta] enable metricbeat upgrade test

jmlrt:
This commit enable Metricbeat upgrade test which where disable before
7.10.0 release because Metricbeat 7.10.0 introduce breaking change
preventing upgrade from previous version.

Version post 7.10.0 will now be tested for upgrade from 7.10.0.


Conky5 (approved):
LGTM

PR #960, opened 2020-12-01
Fix - Metricbeat default values - output module config

njgibbon:
I believe there is a typo in the current configuration. The environment variable is meant to be templated as a prefix in the elastic output 'hosts' value string but the lines misplace the closing brace.

- [x] README.md updated with any new values or changes - n/a.
- [ ] Updated template tests in `${CHART}/tests/*.py` - see comment.
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml` - see comment.

I can't quite work out how to run the template tests and haven't tried with the integration ones.

I think the template tests should check that '${ELASTICSEARCH_HOSTS}' is includes in the output config.

I'd imagine you could check that the host value doesn't get applied as "" at runtime. I don't know how this passes testing now because surely the ES output will never work with the default values. 

Maybe I am missing something aha. Please get back to me on if there is any misunderstanding here or any docs so I can help add to the testing in the ways i've described above.

### DIFF
--- a/filebeat/values.yaml
+++ b/filebeat/values.yaml
@@ -68,7 +68,7 @@ livenessProbe:
       - |
         #!/usr/bin/env bash -e
         curl --fail 127.0.0.1:5066
-  failureThreshold: 3
+  failureThreshold: 5
   initialDelaySeconds: 10
   periodSeconds: 10
   timeoutSeconds: 5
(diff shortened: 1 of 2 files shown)


jmlrt (changes requested):
Hi @njgibbon,
This is not a typo but the expected syntax so that if `ELASTICSEARCH_HOSTS` variable is defined, `hosts` will match it, otherwise, `hosts` will default to `elasticsearch-master:9200`.

nickgibbon-nationwide:
Ah ok I see. There you go \/. I never knew 👍 

https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

Will close this.

njgibbon:
Ah ok I see. I never knew 👍

https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

Will close this.

PR #969, opened 2020-12-09
[logtash] use Helm recommended labels

ebuildy:

**This update labels to Helm recommended format** , according https://helm.sh/docs/chart_best_practices/labels/. I noticed elastic use legacy format when I setup networkPolicy for my elasticsearch cluster.

If this PR is ok, I can work on other elastic products.

### DIFF
--- a/logstash/templates/_helpers.tpl
+++ b/logstash/templates/_helpers.tpl
@@ -44,6 +44,6 @@ helm.sh/chart: {{ include "logstash.chart" . }}
 Labels to use on deploy.spec.selector.matchLabels and svc.spec.selector
 */}}
 {{- define "logstash.matchLabels" -}}
-app.kubernetes.io/name: {{ include "logstash.name" . }}
+app.kubernetes.io/name: {{ include "logstash.fullname" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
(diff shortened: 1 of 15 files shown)


jmlrt:
Hi @ebuildy,
Thank you for submitting this PR. While I agree that we should use labels recommended by Helm, we need to ensure that this doesn't break chart upgrade. I'll try to take a look soon.

jmlrt:
jenkins test this please

ebuildy:
yes, maybe we could keep legacy ones during some times?

jmlrt:
Hi @ebuildy,

So after a bit more investigation, we should wait until next major release (8.0.0) to release this kind of changes.

Our branching model would allow us to merge this change in master without backporting it to release branches to ensure it will nor be released before 8.0.0.

However, for ease of maintainance when backporting other PRs and answering to bug report, I would prefer not merging this PR yet and rework the labels when we'll start working preparing 8.0.0 release (no date estimation for now).

I hope that's fine with you.

jmlrt:
still valid

jmlrt:
still valid

jmlrt:
still valid

tewfik-ghariani:
Will this change ever see the light? ^^

PR #970, opened 2020-12-10
[all] add hostaliases

### LINKED ISSUE #840: Support for hostAliases
Hello,

My output has ssl => true, a certificate, and a host with an IP address. The host gave me a self-signed cert with a peer name such as 'elastic-server'. 

I would rather not use ssl_certificate_verfication => false. Is it possible to get an update to values.yaml to add hostAliases? Is there a way to modify hostAliases without modifying the helm chart?

Thanks

jmlrt:
This commit allow to define hostAliases in all charts

https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/

Fix #840


Conky5 (approved):
LGTM

PR #971, opened 2020-12-10
[meta] add config for backport

jmlrt:
This commit add the config file to manage backport PR using
https://github.com/sqren/backport

### DIFF
--- a/.backportrc.json
+++ b/.backportrc.json
@@ -1,0 +1,8 @@
+{
+  "upstream": "elastic/helm-charts",
+  "targetBranchChoices": ["6.8", "7.10", "7.x"],
+  "all": true,
+  "prFilter": "label:need-backport",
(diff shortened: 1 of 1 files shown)


sqren (approved):
lgtm 👍 
lmk how it goes.

Conky5 (approved):

jmlrt:
Admin merging as the transient issues are not related.

jmlrt:
@sqren thanks for your help and review.

Can you tell me if we need to backport this PR so `.backportrc.json` exists in `6.8`, `7.10` and `7.x` branches or is it enough to have it in `master` branch?

sqren:
While not required it's a good idea to have it on any branch a developer might checkout locally. 
This is because when running the backport script it will look for the config locally, do if a user is on a branch without a config it won't work.

PR #994, opened 2020-12-18
[elasticsearch][kibana] Add flexible ingress

kevinsmithwrs:
- [x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [x] README.md updated with any new values or changes
- [x] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

Fix #979

Issue #536, opened 2020-03-23
[kibana/elasticsearch] ingress defined with hosts breaks other ingress with nginx


aitorhh:
What I would like to suggest is to have support through values configuration for non-hosts definition when deploying kibana (or elasticsearch).

**Chart version:**
```
version: 5.0.11
appVersion: 7.6.1
```

**Kubernetes version:**

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```

**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
Local deployment

**Helm Version:**

```
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

**Describe the bug:**

When creating an ingress with hosts, it makes all others ingress with default "*" to be not accessible. 
I have seen this behaviour with nginx-ingress (stable/nginx-ingress from https://helm.nginx.co/stable appVersion: 1.6.3)

**Steps to reproduce:**

1. `helm install nginx-ingress stable/nginx-ingress`
2. `helm install kibana elastic/kibana --set ingress.enabled=true --set ingress.hosts[0]=test.com`

**Expected behavior:**

I expect that all the ingress should be working regardless

**Any additional context:**
I understand that this issue comes from nginx, but I also believe that it has less impact by updating the helm chart here.

aitorhh:
Until it has being solved in another issues or PR, I believe it's something quite relevant that shouldn't be closed.

jmlrt:
still valid

jmlrt:
still valid

jmlrt:
still valid

jmlrt:
still valid

kloudblog:
anyone facing issue with kibana ingress?
[...]

PR #1003, opened 2021-01-05
[elasticsearch] Make ingress hosts optional

wouterdegeus:
This PR allows the ingress hosts definition to be made empty, which is useful for POC situations where we really don't care about hostnames, for example in GKE using an internal loadbalancer with container native loadbalancing:
```
service:
  annotations:
    cloud.google.com/neg: '{"ingress": true}'

ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: "gce-internal"
  path: /
  hosts:
```

Without this PR the chart would generate an ingress without rules given above config, which is not accepted by kubernetes.
With this PR it generates a proper ingress template without a hostname tied to it.

- [ x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [ x] README.md updated with any new values or changes
- [x ] Updated template tests in `${CHART}/tests/*.py` 

### DIFF
--- a/elasticsearch/templates/ingress.yaml
+++ b/elasticsearch/templates/ingress.yaml
@@ -30,8 +30,9 @@ spec:
   {{- end }}
 {{- end }}
   rules:
-  {{- range .Values.ingress.hosts }}
-    {{- if $ingressPath }}
+  {{- if .Values.ingress.hosts }}
+    {{- range .Values.ingress.hosts }}
+      {{- if $ingressPath }}
     - host: {{ . }}
       http:
         paths:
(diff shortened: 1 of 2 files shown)


jmlrt:
Hi @wouterdegeus,
Thanks for opening this PR.
We'll try to review it soon. Please be patient.

jmlrt:
Related to #536

jmlrt:
jenkins test this please

jmlrt:
Hi @wouterdegeus,

Since #994, you can create an ingress with empty hosts using the following values:
```yaml
ingress:
  enabled: true
  annotations:
    cloud.google.com/neg: '{"ingress": true}'
  hosts:
    - host: ''
      paths:
        - path: /
```

I tested it and had exactly the same ingress generated as using your PR.

Can you confirm this is working for your use case?

wouterdegeus:
Just tested, this works indeed!

jmlrt:
> Just tested, this works indeed!

Awesome 👍
Closing your PR then.

PR #1012, opened 2021-01-06
[elasticsearch] fix secrets in config example

### LINKED ISSUE #1008: basic auth not being enforced for config example
Am following the [config example](https://github.com/elastic/helm-charts/tree/master/elasticsearch/examples/config) to set basic auth. Have cloned the repo and am following it exactly except for this one change:

```
extraEnvs:
  - name: ELASTIC_PASSWORD
    valueFrom:
      secretKeyRef:
        name: elastic-config-credentials
        key: password
  - name: ELASTIC_USERNAME
    valueFrom:
      secretKeyRef:
        name: elastic-config-credentials
        key: username
```
Had to change this from the example which says [elastic-credentials](https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/config/values.yaml#L9) but it must mean [elastic-config-credentials](https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/config/Makefile#L13) as that is the name of the secret. Otherwise I get `Error: secret "elastic-credentials" not found`.

With that change I can install everything.

The example curl of `curl -u elastic:changeme http://localhost:9200/_cat/indices` does indeed give a 200 response. But so does `curl http://localhost:9200/_cat/indices`. It doesn't force me to supply the user and password.

jmlrt:
This commit fix the secret name for credentials in config example

Fix #1008

### DIFF
--- a/elasticsearch/examples/config/test/goss.yaml
+++ b/elasticsearch/examples/config/test/goss.yaml
@@ -2,6 +2,8 @@ http:
   http://localhost:9200/_cluster/health:
     status: 200
     timeout: 2000
+    username: "{{ .Env.ELASTIC_USERNAME }}"
+    password: "{{ .Env.ELASTIC_PASSWORD }}"
     body:
       - "green"
       - '"number_of_nodes":1'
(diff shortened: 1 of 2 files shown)


Conky5 (dismissed):
LGTM

Conky5 (approved):

PR #1018, opened 2021-01-14
[meta] fix transient errors with stable repository

jmlrt:
This commit move stable chart repo setup and dependency update steps to
metricbeat specific examples.

We have lot of transient issues when running configuring stable repo:

```
helm repo add stable https://charts.helm.sh/stable && \
helm dependency update && \
cd ./examples/oss && \
make
Error: looks like "https://charts.helm.sh/stable" is not a valid chart repository or cannot be reached: read tcp 172.17.0.2:37660->185.199.108.153:443: read: connection reset by peer
make: *** [integration] Error 1
```

As stable repo is only used by Metricbeat chart, by not configuring it
for other charts tests, we should limit the risk of transient issues.


Conky5 (approved):
LGTM!

PR #1033, opened 2021-01-27
[meta] add build status and artifact hub badges

jmlrt:
This commit is adding Artifact Hub badges to main and charts README
pages. Also add build status badges to charts README pages (was only
present on main README page).


Conky5 (approved):
LGTM

Issue #733, opened 2020-07-16
[metricbeat] Follow migration from kube-state-metrics charts out of stable repo


jmlrt:
Metricbeat chart currently use [kube-state-metrics](https://github.com/helm/charts/tree/master/stable/kube-state-metrics) chart as subchart to get metrics from K8S.

As [Stable repo](https://github.com/helm/charts/tree/master/stable/) will be deprecated in November 2020 (see [Deprecation timeline](https://github.com/helm/charts#deprecation-timeline)), kube-state-metrics chart will need to migrate to another repository.

This work seem to be tracked in https://github.com/kubernetes/kube-state-metrics/issues/1153.

When the migration will be over, Metricbeat chart will need to be updated to get kube-state-metrics subchart from its new home.

jmlrt:
While waiting for kube-state-metrics chart migration to a new GitHub repository, We need to update Metricbeat chart requirements to retrieve the subchart from the new stable Helm repos url => #882

jmlrt:
still valid

jmlrt:
For later reference, kube-state-metrics chart is now hosted at https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics

christophercutajar:
While deploying an ECK cluster using terraform, I've deployed `kube-state-metrics` v2.13.0 using the official chart from https://kubernetes.github.io/kube-state-metrics. One breaking change I've encountered in the official helm chart is in below line: https://github.com/elastic/helm-charts/blob/0ff3a3c348b2cfe4fd65a17717bf8fc3e15c466d/metricbeat/templates/deployment.yaml#L121

As a solution, I've changed the aforementioned line to read as follows: `"$(KUBE_STATE_METRICS_SERVICE_HOST):$(KUBE_STATE_METRICS_SERVICE_PORT_HTTP)"`

Metricbeat deployment was deployment successfully with this change. 

Happy to help in fully testing and performing the necessary changes!

jmlrt:
still valid

jmlrt:
still valid

christophercutajar:
@jmlrt let me know how can I help in achieving this!

jmlrt:
Hi @christophercutajar, 
Sorry I totally missed your previous comment.
Thanks for your help!

I didn't try the new kube-state-metrics yet. It seems you did it. 

Do you think you could create a PR?

[...]

PR #1039, opened 2021-02-02
[metricbeat] Fixing the respository of kube-state-metrics for metricbeats

yousafsyed:
Kube state metrics that were used in the metricbeat's charts were pointing towards the deprecated helm chart repository i.e "@stable" and in order to fix that we needed to point the chart towards @bitnami/kube-state-metrics which are stable and compatible with the metricbeat's deployment.

Without this small change, the charts are rendered useless since it breaks at the requirements phase.

- [x ] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [✔️ ] README.md updated with any new values or changes
- [x ] Updated template tests in `${CHART}/tests/*.py` 
- [x ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

### DIFF
--- a/metricbeat/README.md
+++ b/metricbeat/README.md
@@ -50,8 +50,7 @@ See [supported configurations][] for more details.
 * Install it:
   - with Helm 3: `helm install metricbeat elastic/metricbeat`
   - with Helm 2 (deprecated): `helm install --name metricbeat elastic/metricbeat`
-
-Note: If you choose to enable `kube-state-metrics` than you should also add the bitnami repostory to your helm3 installation as follows
+  - Add the Elastic Helm charts repo (required for kube-state-metrics chart dependency): `helm repo add stable https://charts.helm.sh/stable`
 
 ```
 helm repo add bitnami https://charts.bitnami.com/bitnami
--- a/metricbeat/README.md
+++ b/metricbeat/README.md
@@ -48,9 +48,9 @@ See [supported configurations][] for more details.
 `helm repo add elastic https://helm.elastic.co`
 
 * Install it:
+  - Add the Elastic Helm charts repo (required for kube-state-metrics chart dependency): `helm repo add stable https://charts.helm.sh/stable`
   - with Helm 3: `helm install metricbeat elastic/metricbeat`
   - with Helm 2 (deprecated): `helm install --name metricbeat elastic/metricbeat`
-  - Add the Elastic Helm charts repo (required for kube-state-metrics chart dependency): `helm repo add stable https://charts.helm.sh/stable`
 
 ### Install development version using master branch
 
--- a/metricbeat/README.md
+++ b/metricbeat/README.md
@@ -51,6 +51,12 @@ See [supported configurations][] for more details.
   - with Helm 3: `helm install metricbeat elastic/metricbeat`
   - with Helm 2 (deprecated): `helm install --name metricbeat elastic/metricbeat`
 
+Note: If you choose to enable `kube-state-metrics` than you should also add the bitnami repostory to your helm3 installation as follows
+
+```
+helm repo add bitnami https://charts.bitnami.com/bitnami
+```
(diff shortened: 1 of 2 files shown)


jmlrt:
Hi @yousafsyed,

The current `@stable` requirements is still working as long as you configured stable repo to target the new location using `helm repo add stable https://charts.helm.sh/stable`.

This is a temporary solution as https://charts.helm.sh/stable is not anymore updated but this is fine for now as we pinned kube-state-metrics 1.2.0 for purpose.

We had a few bad experiences before where bumping to new kube-state-metrics version broke metricbeat chart and we won't update it until we have time to fully test it.

For reference, the work to update kube-state-metrics dependency is already tracked in https://github.com/elastic/helm-charts/issues/733 and we intend to use the official kube-state-metrics chart from https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics instead of bitnami chart.

**What can be improved for now?**

With Helm 2, stable repo was configured by default when installing Helm client, so there was no need to document it.

With Helm 3, stable repo is no more configured by default and it's URL changed to https://charts.helm.sh/stable.

We need to explicitely mention it in the [chart install section](https://github.com/elastic/helm-charts/blob/master/metricbeat/README.md#installing), with something like:

>* Add the Elastic Helm charts repo (required for [kube-state-metrics chart](https://github.com/helm/charts/tree/master/stable/kube-state-metrics) dependency):
>`helm repo add stable https://charts.helm.sh/stable`

Would you be interested to update your PR to add this instruction instead?

jmlrt (changes requested):
Hi @yousafsyed,

The current `@stable` requirements is still working as long as you configured stable repo to target the new location using `helm repo add stable https://charts.helm.sh/stable`.

This is a temporary solution as https://charts.helm.sh/stable is not anymore updated but this is fine for now as we pinned kube-state-metrics 1.2.0 for purpose.

We had a few bad experiences before where bumping to new kube-state-metrics version broke metricbeat chart and we won't update it until we have time to fully test it.

For reference, the work to update kube-state-metrics dependency is already tracked in https://github.com/elastic/helm-charts/issues/733 and we intend to use the official kube-state-metrics chart from https://github.com/kubernetes/kube-state-metrics/tree/master/charts/kube-state-metrics instead of bitnami chart.

**What can be improved for now?**

With Helm 2, stable repo was configured by default when installing Helm client, so there was no need to document it.

With Helm 3, stable repo is no more configured by default and it's URL changed to https://charts.helm.sh/stable.

We need to explicitely mention it in the [chart install section](https://github.com/elastic/helm-charts/blob/master/metricbeat/README.md#installing), with something like:

>* Add the Elastic Helm charts repo (required for [kube-state-metrics chart](https://github.com/helm/charts/tree/master/stable/kube-state-metrics) dependency):
>`helm repo add stable https://charts.helm.sh/stable`

Would you be interested to update your PR to add this instruction instead?

yousafsyed:
That makes more sense

yousafsyed:
@jmlrt
I guess this is fine for now, It's working for me and I have updated the documentation instead  of the change that I suggested earlier although the bitnami charts for kube-state-metric worked for me, but I agree we need to test them well before changing the version or charts.

metricbeat/README.md
```diff
@@ -50,7 +50,7 @@ See [supported configurations][] for more details.
 * Install it:
   - with Helm 3: `helm install metricbeat elastic/metricbeat`
   - with Helm 2 (deprecated): `helm install --name metricbeat elastic/metricbeat`
-
+  - Add the Elastic Helm charts repo (required for kube-state-metrics chart dependency): `helm repo add stable https://charts.helm.sh/stable`
```
jmlrt, line 53: NIT but I would put it just before the `helm install` commands because the install will fail if it's not done before.

yousafsyed: applied the changes

jmlrt (approved):
LGTM⛴

PR #1042, opened 2021-02-03
Fix formatting issue in `values.yaml`

fmd:
The template looks like this:

```
    {{- if $ingressPath }}
    - host: {{ . }}
      http:
        paths:
          - path: {{ $ingressPath }}
            backend:
              serviceName: {{ $fullName }}
              servicePort: {{ $httpPort }}
```

where `$ingressPath` is `.Values.ingress.path`

This means that following the format in the `values.yaml` doesn't work. Assuming that the template is correct, this is the fix (it works for us)

### DIFF
--- a/elasticsearch/values.yaml
+++ b/elasticsearch/values.yaml
@@ -244,9 +244,8 @@ ingress:
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
   hosts:
-    - host: chart-example.local
-      paths:
-        - path: /
+    - chart-example.local
+  path: /
   tls: []
   #  - secretName: chart-example-tls
   #    hosts:


fmd:
I've signed the CLA but for some reason it's saying I havent

jmlrt:
Hi @fmd,
Flexible ingress have been added in #994 and using `$ingressPath` is now the deprecated way that we kept only for backward compatibility.

You can see in the ingress template code that `$ingressPath` is only used when `.Values.ingress.path` exist, otherwise it use the new way:

https://github.com/elastic/helm-charts/blob/e07138cd2ceb3f2dc97b9d162ec09d00f1f22fcc/elasticsearch/templates/ingress.yaml#L34-L52

PR #1074, opened 2021-02-22
[elasticsearch] Add volumeClaimTemplate.storageClassName

RammusXu:
default to ssd

### DIFF
--- a/elasticsearch/values.yaml
+++ b/elasticsearch/values.yaml
@@ -97,6 +97,7 @@ sidecarResources: {}
 networkHost: "0.0.0.0"
 
 volumeClaimTemplate:
+  # storageClassName: ssd
   accessModes: [ "ReadWriteOnce" ]
   resources:
     requests:


jmlrt:
cla/check

jmlrt (changes requested):
Hi @RammusXu, can you be a bit more explicit about why we should add this commented value? 

Also signing the CLA is required for PR: https://www.elastic.co/contributor-agreement

RammusXu:
> Hi @RammusXu, can you be a bit more explicit about why we should add this commented value?

When I try to deploy this to alibabacloud hosted kubernetes, it fails. I need to specify the `storageClassName` to make it run. 
Thus, I think `storageClassName: ssd` will be helpful.

Also, I can see it as default values anywhere, like https://github.com/bitnami/charts/blob/master/bitnami/redis-cluster/values.yaml#L705-L708

> Also signing the CLA is required for PR: https://www.elastic.co/contributor-agreement

I have signed it and now I can see the CLA check is green. Is it ok?

jmlrt:
still valid

PR #1089, opened 2021-03-03
[elasticsearch] only configure ES_JAVA_OPTS when value is set

### LINKED ISSUE #1083: Can we disable the  default setting of JVM heap size in statefulset.yaml?
[Starting from version 7.11, Elasticsearch automatically sizes JVM heap based on a node’s roles and total memory.](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-jvm-heap-size.html) 
With this higher-order version,  someone don't want to set the  default setting of JVM heap size,  the corresponding templates needs to be configured.
from `elasticsearch/templates/statefulset.yaml`
``` 
318           - name: ES_JAVA_OPTS
319             value: "{{ .Values.esJavaOpts }}"
```
Change the configuration to this below maybe work:
```
{{- if .Values.esJavaOpts  }}
          - name: ES_JAVA_OPTS
            value: "{{ .Values.esJavaOpts }}"
{{- end }}
```
With the higher-order version, we can overwrite  `esJavaOpts` to null in custom configrations.

jmlrt:
This commit update Elasticsearch Statefulset template to set
`ES_JAVA_OPTS` variable only when `esJavaOpts` value is defined.

This allow to benefit of JVM heap automatic sizing in Elasticsearch
7.11.

Fix #1083

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -315,8 +315,10 @@ spec:
             value: "{{ .Values.clusterName }}"
           - name: network.host
             value: "{{ .Values.networkHost }}"
+          {{- if .Values.esJavaOpts  }}
           - name: ES_JAVA_OPTS
             value: "{{ .Values.esJavaOpts }}"
+          {{- end }}
           {{- range $role, $enabled := .Values.roles }}
           - name: node.{{ $role }}
             value: "{{ $enabled }}"
--- a/elasticsearch/values.yaml
+++ b/elasticsearch/values.yaml
@@ -68,7 +68,7 @@ podAnnotations: {}
 # additionals labels
 labels: {}
 
-esJavaOpts: "-Xmx1g -Xms1g"
+esJavaOpts: "" # example: "-Xmx1g -Xms1g"
 
 resources:
   requests:
--- a/elasticsearch/tests/elasticsearch_test.py
+++ b/elasticsearch/tests/elasticsearch_test.py
@@ -45,7 +45,6 @@ def test_defaults():
         {"name": "discovery.seed_hosts", "value": uname + "-headless"},
         {"name": "network.host", "value": "0.0.0.0"},
         {"name": "cluster.name", "value": clusterName},
-        {"name": "ES_JAVA_OPTS", "value": "-Xmx1g -Xms1g"},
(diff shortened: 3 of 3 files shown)


mehta-ankit (dismissed):

elasticsearch/values.yaml
```diff
@@ -68,7 +68,7 @@ podAnnotations: {}
 # additionals labels
 labels: {}
 
-esJavaOpts: "-Xmx1g -Xms1g"
+esJavaOpts: "" # example: "-Xmx1g -Xms1g"
```
Conky5, line 71: I'm not sure this correctly formatted, should the whole `esJavaOpts` line be commented out to have it not be set by default?

elasticsearch/templates/statefulset.yaml
```diff
@@ -315,8 +315,10 @@ spec:
             value: "{{ .Values.clusterName }}"
           - name: network.host
             value: "{{ .Values.networkHost }}"
+          {{- if .Values.esJavaOpts  }}
```
Conky5, line 318: It looks like there is a test asserting the default ES_JAVA_OPTS option that will need updated https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+template-testing/1218/CHART=elasticsearch,label=docker&&virtual/console

Conky5 (approved):
LGTM

viceice:
This broke my upgrade from `6.8.16` to `6.8.17` with constantly crashing elastice search v6

jmlrt:
Hi @viceice, the upgrade from 6.8.16 to 6.8.17 with this feature was tested successfully in our CI.
Can you open a new issue with all the details to understand what problem did you add?

viceice:
@jmlrt I'm using `imageTag: docker.elastic.co/elasticsearch/elasticsearch:5.6.16`, so it's probably a problem with the ancient es version. I can't currently switch to a newer es, because of some breakings i need so solve first.

jmlrt:
Hi @viceice, without much investigation I think this may be related to the old Elasticsearch version.
Note that only 6.8.X and 7.X versions are supported with these Helm charts. 
All previous versions were never tested by Elastic with these charts and we don't try to maintain any compatibility.

PR #1108, opened 2021-03-17
[elasticsearch] skip setting the node.remote_cluster_client env for ES6

karolinepauls:
- [ v ] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [ v ] README.md updated with any new values or changes
- [ v ] Updated template tests in `${CHART}/tests/*.py` 


karolinepauls:
https://www.elastic.co/guide/en/elasticsearch/reference/6.8/modules-node.html - missing in ES6.

jmlrt (changes requested):
Hey @karolinepauls, Elasticsearch 6.x charts are using the `6.8` git branch which doesn't include `remote_cluster_client` role.

Note that we plan to remove the `esMajorVersion` and related conditions later as only charts from `6.8` branch or `6.8.x` git tags should be used with Elasticsearch 6.x.

karolinepauls:
Yes, I've just noticed and switched to 6.8...

If esMajorVersion is going away, I suppose it doesn't make sense to extract the test from this PR then.

But it makes sense to add information that esMajorVersion is deprecated.

PR #1109, opened 2021-03-17
[elasticsearch] Mark esMajorVersion as deprecated

karolinepauls:


jmlrt (approved):
LGTM⛴

PR #1111, opened 2021-03-17
[elasticsearch] fix network policies http additional rules

### LINKED ISSUE #1106: Network Policy - Additonnal Rule
Chart version: 7.12

Describe the bug:

In the NetworkPolicy definition, `.Values.networkPolicy.transport.additionalRules` is used in both the `http` and `transport` blocks. The http block should use `.Values.networkPolicy.http.additionalRules` as defined in the `values.yaml` file.

Problematic line:
https://github.com/elastic/helm-charts/blob/90f71e7faa1bd838ce8aff9455e9acae07809625/elasticsearch/templates/networkpolicy.yaml#L31

jmlrt:
* Fix http additional rules to use with `NetworkPolicies`.
* Adds some small improvements in network policies tests:
  * update example Makefile to fix commands and use similar structure to other examples Makefiles
  * fix example values file extension
  * use different values for http and transport network policies in python test and example values to catch regressions where same values would be used for both

Fix #1106

### DIFF
--- a/elasticsearch/examples/networkpolicy/Makefile
+++ b/elasticsearch/examples/networkpolicy/Makefile
@@ -1,13 +1,14 @@
 default: test
+
 include ../../../helpers/examples.mk
 
 RELEASE := helm-es-networkpolicy
+TIMEOUT := 1200s
 
 install:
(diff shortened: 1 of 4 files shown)


Conky5 (approved):
LGTM

Conky5 (approved):
LGTM

PR #1116, opened 2021-03-18
[elasticsearch] add value to disable tests

nflaig:
It should be possible to disable the tests of the chart. This is especially important if the chart is used as a subchart which means the test are also executed if you run `helm test`. For some users it might not be desirable to run the test provided by the elastic helm chart, the value gives the user of the chart the option to configure this.


jmlrt (commented):
Thanks for this PR, can you update the README with this new value?

nflaig:
@jmlrt Thanks again for the review! I updated the README with the new value

nflaig:
@jmlrt any update on this? Let me know if I should apply further changes

jmlrt (approved):
Fixed the conflict. Change LGTM. As a minor improvement, I think it could be great to also add a `{{- if .Values.tests.enabled -}}` condition to not print test command when the test is disabled.

https://github.com/elastic/helm-charts/blob/0e5a168fdcdab22cb9ed2f89e5ccb79e6c4d82fd/elasticsearch/templates/NOTES.txt#L3-L4

jmlrt (approved):
LGTM⛴

jmlrt:
jenkins test this please

PR #1119, opened 2021-03-23
Remove slack notification

Conky5:
Remove the CI slack notification, these tests are too flaky and create a
lot of non-actionable noise in this notification room.

- ~[ ] Chart version *not* bumped (the versions are all bumped and released at the same time)~
- ~[ ] README.md updated with any new values or changes~
- ~[ ] Updated template tests in `${CHART}/tests/*.py`~
- ~[ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`~

### DIFF
--- a/.ci/jobs/elastic+helm-charts+master.yml
+++ b/.ci/jobs/elastic+helm-charts+master.yml
@@ -46,10 +46,3 @@
       - project: elastic+helm-charts+master+cluster-cleanup
         current-parameters: true
         trigger-with-no-params: false
-    - slack:
-        notify-back-to-normal: True
-        notify-every-failure: True
-        room: infra-release-notify
-        team-domain: elastic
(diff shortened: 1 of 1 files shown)


jmlrt (approved):
LGTM, I added these notifications when we had the "infra-release oncall" to triage build failures in the hope that everyone in the team would look at them during their oncall and re-run jobs for transient issues / investigate a bit or ping me for real issues, but:

1. we don't have anymore the "infra-release oncall"
2. I'm the only one to take a look at these failures even when we had the "infra-release oncall" and I don't need these notifications to remember to take a look regularly
3. fixing these transient issues is not a priority so I don't expect jobs to be less flaky in short term.

Conky5:
Thanks for the information @jmlrt , I was thinking that is the situation we're in currently. If one of these things changes I think we can definitely add these notifications back.

Issue #1035, opened 2021-01-29
Support for elastalert plugin


dcvtruong:
**Describe the feature:**
```
Add elastalert plugin via helm.
```
**Describe a specific use case for the feature:**

ebuildy:
Please read: https://github.com/elastic/helm-charts/tree/master/elasticsearch#how-to-install-plugins

> The recommended way to install plugins into our Docker images is to create a custom Docker image.

Issue #869, opened 2020-10-22
ElasticSearch: Allow to install plugins via values.yaml


clausa20:
Old unofficial helm allowed to install plugins specifying a list in the values.yaml (cluster: plugins: []).
Now the official helm does not allow it, recommending to use a custom Dockerfile.
Using a custom Dockerfile seems not to be a good idea, because it forces to fork the original and stable ES image and it is harder to maintain in a separated manner.

You can use as old helm, an init-container that installs the necessary plugins when the pod is starting.

ebuildy:
Can you share old Helm chart URL please?

Also, about "custom Dockerfile seems not to be a good idea" , this is the best practice recommanded by Docker it-self, the idea behind container is immutability! 

It's also a good practice to package OS/Softwares/Config/App code,vendors within the Docker image.

A (good) production infrastructure should never allow a container to make Internet call, for security reason, in the case of Elasticsearch plugin, imagine if someone add bad code in the plugin JAR, or even, if S3/Github is not available when you spaw new pod!

PR #1137, opened 2021-04-11
[elasticsearch] New Feature - Adding plugin support through values file

grantholly:

I've seen a couple of issues (https://github.com/elastic/helm-charts/issues/1035, https://github.com/elastic/helm-charts/issues/869) created around wanting to support adding plugin support through the values file.  I had to figure this out myself, and so I thought I would submit this change back for review as a new feature.  

I totally understand the recommendation of using a custom container image with your plugins baked in: more immutability, no dependency on startup to fetch plugins, etc., but sometimes using a custom image is not an issue because you don't have a private image repository for your use.


jmlrt:
Hi @grantholly, thanks for this PR, we may revisit it one day but for now, we still prefer not adding this feature in the charts for the known reliability reasons.

Maybe you can deploy the registry as a pod in your K8S cluster for these kind of use cases.

Issue #1138, opened 2021-04-11
[apm-server] service could not bind to pod when use 'fullnameOverride'


ghilbut:
**Chart version:**
7.12.0

**Kubernetes version:**
1.19.7

**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
docker-desktop

**Helm Version:**
v3

**`helm get release` output**

e.g. `helm get elasticsearch` (replace `elasticsearch` with the name of your helm release)

*Be careful to obfuscate every secrets (credentials, token, public IP, ...) that could be visible in the output before copy-pasting.*

*If you find some secrets in plain text in `helm get release` output you should use [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to managed them is a secure way (see [Security Example](https://github.com/elastic/helm-charts/blob/master/elasticsearch/examples/security/values.yaml#L23-L38)).*

<details>
<summary>Output of helm get release</summary>

```
```

</details>

**Describe the bug:**

**Steps to reproduce:**

1.
2.
3.

**Expected behavior:**

**Provide logs and/or server output (if relevant):**

*Be careful to obfuscate every secrets (credentials, token, public IP, ...) that could be visible in the output before copy-pasting*

```
```

**Any additional context:**

APM-Server service could not bind to pod when use 'fullnameOverride'. Because of deployment's 'spec.selector.matchLabels.app' and 'spec.template.metadata.labels.app' field values are hard coding to 'apm-server'. It should be better to change to '{{ .Chart.Name }}' variable.

PR #1139, opened 2021-04-11
[apm-server] fix service could not bind to pod

ghilbut:
Fix https://github.com/elastic/helm-charts/issues/1138

### DIFF
--- a/apm-server/templates/deployment.yaml
+++ b/apm-server/templates/deployment.yaml
@@ -14,12 +14,12 @@ spec:
 {{ toYaml .Values.updateStrategy | indent 4 }}
   selector:
     matchLabels:
-      app: apm-server
+      app: {{ .Chart.Name }}
       release: {{ .Release.Name | quote }}
   template:
     metadata:
       labels:
-        app: apm-server
+        app: {{ .Chart.Name }}
         release: {{ .Release.Name | quote }}
       annotations:
         {{- range $key, $value := .Values.podAnnotations }}


apm-server/templates/deployment.yaml
```diff
@@ -14,12 +14,12 @@ spec:
 {{ toYaml .Values.updateStrategy | indent 4 }}
   selector:
     matchLabels:
-      app: apm-server
+      app: {{ .Chart.Name }}
```
jmlrt, line 17:
```suggestion
      app: "{{ template "apm.fullname" . }}"
```

apm-server/templates/deployment.yaml
```diff
@@ -14,12 +14,12 @@ spec:
...
       release: {{ .Release.Name | quote }}
   template:
     metadata:
       labels:
-        app: apm-server
+        app: {{ .Chart.Name }}
```
jmlrt, line 22:
```suggestion
        app: "{{ template "apm.fullname" . }}"```

jmlrt (commented):
Thanks for this PR 👍🏻 
This LGTM but I think we could use [`template "apm.fullname"`](https://github.com/elastic/helm-charts/blob/2c9fceb6b8bcff05d06a0c05aae56957994c1319/apm-server/templates/_helpers.tpl#L13) like in Filebeat and Metricbeat charts instead. WDYT?

Issue #1150, opened 2021-04-14
Allow using templates to define secrets in logstash chart


stuart-mclaren-hpe:
**Describe the feature:**

It is currently possible to define secrets in the logstash helm chart.
You have two options: (1) a key/value pair (2) reference a secret in a file.

I'd like to add a third option where the value can be interpreted as a template.

In code, this would be a couple of additional lines (in secrets.yaml):

```
data:                                                                           
{{- range $key, $val := .value }}                                               
  {{- if hasSuffix "filepath" $key }}                                           
  {{ $key | replace ".filepath" "" }}: {{ $.Files.Get $val | b64enc | quote }}  
  {{ else if hasSuffix "template" $key }}                                                        <<< new line
  {{ $key | replace ".template" "" }}: {{ tpl $val $ | b64enc | quote }}             <<< new line
  {{ else }}                                                                    
  {{ $key }}: {{ $val | b64enc | quote }}                                       
  {{- end }}                                                                    
{{- end }} 
```

**Describe a specific use case for the feature:**

We use the logstash chart as a dependent chart. This means we have a values.yaml file which "wraps" the values.yaml in the upstream logstash chart. We'd like to provide a simple way to propagate
a value to a secret that will exist in all our deployments.

The values.yaml in our parent chart would look like:

```
  logstash:
    secrets:                                                                      
      - name: logz-token                                                          
        value:                                                                    
          LOGZ_TOKEN.template: "{{ .Values.logstash.token }}"   
```

Consumers of our chart can then simply provide the token value, without having to worry about the specific secret structure (this avoids passing an array index number etc).

`helm install foo bar --set token=secret123`

For example, you could then do

```
  envFrom:                                                                      
    - secretRef:                                                                
        name: logz-token 
```

to have the environment variable in a pod set to the relevant value:

```
LOGZ_TOKEN=secret123
```

stuart-mclaren-hpe:
Proposed patch here:

https://github.com/elastic/helm-charts/pull/1151

PR #1151, opened 2021-04-14
logstash secret creation with template                               …

stuart-mclaren-hpe:
…                                                                                     [19/878]

Add the ability to specify a secret value using a template.

When the logstash chart is a dependent chart, this allows the parent chart
to specify a secret's contents by using a value.

For example, the parent chart can be installed with a command such as:

`helm install foo bar --set token=secret123`

rather than having to specify the full structure of the secret.

The logstash chart's values.yaml for the above example would look
similar to:

```
    secrets:
      - name: logz-token
        value:
          LOGZ_TOKEN.template: "{{ .Values.logstash.token }}"
```

Combined with envFrom this creates a simple way for
users of the parent chart to set known environment variables to secret values.

Further details on the use case here:

  https://github.com/elastic/helm-charts/issues/1150


stuart-mclaren-hpe:
>  Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):
> [f277da4](https://github.com/elastic/helm-charts/commit/f277da45477fe28128f7af55c35bea7a43dd8faa)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

I have signed this now.

stuart-mclaren-hpe:
Hi @jmlrt 

Is there anything else I need to do at this point? (eg to kick off devops-ci)

Thanks

PR #1183, opened 2021-05-12
[elasticsearch] remove masterTerminationFix

jmlrt:
This PR removes the `masterTerminationFix` side-car container
introduced in #63 to fix slow elections issues when master node is
deleted. This workaround is no more needed since Elasticsearch 7.2.

Also fix values table formatting in README.


Conky5 (approved):
LGTM

PR #1184, opened 2021-05-12
[meta] remove gke 1.16 tests

jmlrt:
This commit remove tests on GKE 1.16 following new GKE version updates:

https://cloud.google.com/kubernetes-engine/docs/release-notes#May_04_2021


Conky5 (approved):
LGTM

PR #1185, opened 2021-05-12
[elasticsearch] remove unused sidecarResources value

jmlrt:
This commit is a follow-up of #1183 to remove the sidecarResources value
which is no more used in the chart.


Conky5 (approved):
LGTM

PR #1196, opened 2021-05-24
[kibana] add emptyDir volume for kibana data

ebuildy:

Use an emptyDir to store Kibana data instead of Docker overlay, this is a good practice, by setting security context  ``readOnlyRootFilesystem`` : true .

User can set securityContext via ``value.yaml`` file.

(we do that for every Helm releases, elasticsearch/APM/logstash and Kibana, so I can create PR for each if this is a good idea of course).

### DIFF
--- a/kibana/templates/deployment.yaml
+++ b/kibana/templates/deployment.yaml
@@ -40,6 +40,8 @@ spec:
       hostAliases: {{ toYaml .Values.hostAliases | nindent 6 }}
       {{- end }}
       volumes:
+        - name: kibana-data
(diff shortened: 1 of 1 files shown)


jmlrt (approved):
LGTM⛴

jmlrt:
jenkins test this please

ebuildy:
actually, with https://github.com/elastic/helm-charts/pull/1247 PR (extraVolumes), I think we dont need this one any more.

Give a values example should be enough

PR #1212, opened 2021-05-25
[7.x] [elasticsearch] only configure ES_JAVA_OPTS when value is set (#1089)

jmlrt:
 - [elasticsearch] only configure ES_JAVA_OPTS when value is set (#1089)


jmlrt:
jenkins test this please

PR #1213, opened 2021-05-25
[6.8] [elasticsearch] remove masterTerminationFix (#1183)

jmlrt:
Backports the following commits to 6.8:
 - [elasticsearch] remove masterTerminationFix (#1183)


jmlrt:
waiting @elastic/es-delivery that this can be backported on 6.8 branch (https://github.com/elastic/helm-charts/pull/1183#issuecomment-839752102)

mark-vieira:
As far as I can tell https://github.com/elastic/elasticsearch/issues/29025 was never backported to 6.8 so I think we'll need to keep this workaround in place. Might want to reach out to @DaveCTurner for more details though.

DaveCTurner:
I admit I don't follow what the code in this PR is doing exactly, but I can confirm that we have not made any changes to 6.8 in this area so https://github.com/elastic/elasticsearch/issues/29025 and https://github.com/elastic/helm-charts/issues/63 still apply to those versions.

jmlrt:
Thanks for your help @mark-vieira @DaveCTurner. Closing this PR as there is no need to backport it to 6.8 then.

PR #1189, opened 2021-05-14
[elasticsearch] fix statefulset to rollout in upgrade test

jmlrt:
This commit fix the elasticsearch statefulset to rollout.

Also update the target version to the last released version

PR #1218, opened 2021-05-25
[7.x] [elasticsearch] fix statefulset to rollout in upgrade test (#1189)

jmlrt:
 - [elasticsearch] fix statefulset to rollout in upgrade test (#1189)


jmlrt:
jenkins test this please

Issue #1226, opened 2021-05-26
[metricbeat] Missing pod security context


ebuildy:
**Chart version:** 7.12.X

The ``deployment`` template is missing pod securityContext, the value variable ``podSecurityContext`` is currently used by the container:

https://github.com/elastic/helm-charts/blob/v7.12.0/metricbeat/templates/deployment.yaml#L129

Like all other chart, we must be able to configure both pod and container securityContext.

lepouletsuisse:
I just face it. IMO this is a bug because PodSecurityContext should be at Pod level, not at container level (as there is already a `deployment.securityContext` key available).
If I receive the approval of the maintainers, I have no problem doing a PR to fix it.

jmlrt:
still valid

PR #1261, opened 2021-06-22
[metricbeat] Fix podSecurityContext bug

lepouletsuisse:
- Fix issue #1226  the `podSecurityContext` value to be compliant with all the other Helm chart about its signification (and also with the README.md that clearly specify this value set the **pod execution security context** and not the container's one)

Note: I tried to fix the tests but I just couldn't make it. I do the PR now so people can use this branch if they absolutely need to set the `podSecurityContext` like me. If anybody else wants to fix the tests, they are welcome to add a commit to this branch.


jmlrt:
cla/check

jmlrt:
jenkins test this please

jmlrt:
still valid, need to find time for review

PR #1264, opened 2021-06-24
[kibana] adding extra volumes and extra volume mounts (#557)

DilasserT:
- [x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [x] README.md updated with any new values or changes
- [x] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

PR Following #1247 

Adding options to mount additionnal volumes with `extraVolumes` and `extraVolumeMounts` through `values.yaml`

@jmlrt kibana/tests/kibana_test.py formatted using `black`

Issue #557, opened 2020-04-03
Make Kibana chart able to acept extraVolumes and extraVolumeMounts


nerddelphi:
Hi there.
It'd be great if **Kibana** chart provide [extraVolume](https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/statefulset.yaml#L130) and [extraVolumeMounts](https://github.com/elastic/helm-charts/blob/master/elasticsearch/templates/statefulset.yaml#L292), like [elasticsearch](https://github.com/elastic/helm-charts/tree/master/elasticsearch) does.

It's very simple and easy and works like a charm with **extraInitContainer** setting.
Thanks.

nerddelphi:
any update?

ados1991:
i agree with @nerddelphi, also any update ?

jmlrt:
Hi @nerddelphi & @ados1991,
Sorry for the late answer.
I don't see a good use case needed mounting additional volumes for Kibana.
Could you give more details why this is needed?

predogma:
We have another customer requesting this.  They basically wanted "...to mount the volume where ES generated certificates for secured ES to Kibana communication."

Angelin01:
> Hi @nerddelphi & @ados1991,
> Sorry for the late answer.
> I don't see a good use case needed mounting additional volumes for Kibana.
> Could you give more details why this is needed?

I am not one of the original posters, however I believe I have a very simple use case: I want to add a sidecar to Kibana, which is already possible through `extraContainers`. Said sidecar might need a volume for whatever reason.

Angelin01:
@jmlrt 
>Hi @nerddelphi & @ados1991,
Sorry for the late answer.
I don't see a good use case needed mounting additional volumes for Kibana.
Could you give more details why this is needed?

I am not one of the original posters, however I believe I have a very simple use case: I want to add a sidecar to Kibana, which is already possible through `extraContainers`. Said sidecar might need a volume for whatever reason.

jmlrt:
Fixed by #557

PR #1280, opened 2021-07-05
[7.x] [kibana] adding extra volumes and extra volume mounts (#557) (#1264)

jmlrt:
 - [kibana] adding extra volumes and extra volume mounts (#557) (#1264)


jmlrt:
jenkins test this please

PR #1294, opened 2021-07-09
[meta] add tests for k8s 1.20

jmlrt:
This commit add tests on GKE 1.20 following new GKE version updates:

https://cloud.google.com/kubernetes-engine/docs/release-notes#June_25_2021


Conky5 (approved):
LGTM

PR #1312, opened 2021-07-30
Support networking.k8s.io/v1 ingress for k8s v1.22+

mkhpalm:
Changes to ingress template to support k8s v1.22+ due to depreciation of `networking.k8s.io/v1beta1`.

* https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

Following the updated boilerplate template structure from helm to keep things consistent

* https://github.com/helm/helm/blob/d4cc130fa90027e8c1c69ad2a06c6dee4c31d29d/pkg/chartutil/create.go#L215-L275

### DIFF
--- a/elasticsearch/templates/ingress.yaml
+++ b/elasticsearch/templates/ingress.yaml
@@ -26,7 +26,10 @@ metadata:
 {{ toYaml . | indent 4 }}
 {{- end }}
 spec:
-{{- if .Values.ingress.tls }}
+  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
+  ingressClassName: {{ .Values.ingress.className }}
+  {{- end }}
+  {{- if .Values.ingress.tls }}
   tls:
   {{- if .ingressPath }}
   {{- range .Values.ingress.tls }}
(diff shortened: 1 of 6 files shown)


maybe-sybr:
@mkhpalm There are also ingress YAML docs for kibana at logstash. Check out #1233 which was a less portable version of your change, but which did touch those documents as well. I'd really like to see one of these land.

mkhpalm:
I've updated the templates for both logstash and kibana in addition to elasticsearch

bender316:
I solved it for the jaeger helm charts with `.Capabilities.APIVersions.Has`:
https://github.com/jaegertracing/helm-charts/pull/275

I think the structure is a bit clearer there and better readable in the future.

mkhpalm:
> I solved it for the jaeger helm charts with `.Capabilities.APIVersions.Has`:
> [jaegertracing/helm-charts#275](https://github.com/jaegertracing/helm-charts/pull/275)
> 
> I think the structure is a bit clearer there and better readable in the future.

I think the problem with this approach is:

1. Only supports k8s versions greater than v1.14
2. Doesn't properly support k8s v1.18 which introduced ingressClassName
3. Misses the point of the v1 spec push to introduce path type abstractions for ingress controller portability
4. Isn't moving people away from the soon to be completely depreciated extensions interface
5. Is a "roll your own" interface that will only be present in jeager with limited control and will be unlike everything else out there

I really do believe helm themselves have the better approach here. If readability is an issue one of the chart maintainers could drop their old ingressPath

ckotzbauer:
Is there any ETA when this is merged? Currently Kibana cannot be installed on Kubernetes 1.22 without changes to the apiserver.

mpepping:
+1 Kubernetes releases are pushing to a point in time that this needs to be merged :)

framsouza:
Closing this PR as we are currently working here #1409

Issue #404, opened 2019-12-09
[Beats] Remove hardcoded or make configurable the docker mounts


mitchellmaler:
Not all Kubernetes clusters deploy using docker since the CRI has multiple implementations. Our situation is we deploy with containerd and not docker and the docker directories or socet do not exist. The beat helm charts have the docker directories mounted by default. 
It would be better to make a configuration to disable mounting the docker directories for clusters that are not using docker at all.

jmlrt:
Hi @mitchellmaler,
Thanks for submitting this feature request.
I'll let the issue open but don't think we'll have time to work on it.

jmlrt:
still valid even if we have no plan to work on it currently :(

jmlrt:
still valid

jmlrt:
still valid

jmlrt:
still valid

f35f0ef9d0e827dae86552d3899f78fc:
This is more relevant now since Azure's Kubernetes (AKS) [switched to using ContainerD](https://github.com/Azure/AKS/releases/tag/2020-11-16) by default as of Kubernetes 1.19.

jmlrt:
still valid

towolf:
still valid

PR #1324, opened 2021-08-05
[7.x][beats] Configurable Docker volumes and volume mounts

charliefoxrf:

This PR attempts to address https://github.com/elastic/helm-charts/issues/404 by adding `daemonset.includeDockerMounts` boolean values for both the `filebeat` and `metricbeat` charts that, when `false`, removes the docker socket and docker container host path volume mounts from the DaemonSets.


v8v5v:
could this issue be re-opened, please?
it looks like an useful feature to have

j771:
This needs to be updated in order for these Helm charts to be compatible with k8s v1.24+ correct?

xnuamx:
Dear colleagues!
Faced the same problem when deploying daemonset filebeat, due to volume varlibdockercontainers and varrundockersock. There is no docker support in the new versions of kybernetes. The solution is simple, add some docker variable and adjust the file https://github.com/elastic/helm-charts/blob/v8.5.1/filebeat/templates/daemonset.yaml

There is work for a couple of minutes, but this problem remains without a solution for many months in the presence of an open issue. Please make the necessary changes!

PR #1337, opened 2021-09-13
[meta] remove support matrix + nit doc changes

jmlrt:
This commit removes the existing support matrix and replace it by a
mention of the charts status (alpha, beta, ga) in the stack version
table. This change will avoid issue where the support matrix isn't
updated during a release.

Also remove the links to stack product docker image doc which are
already present in each chart specific README.

Also reorganize informations related to supported configuration under
this title and format line length.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,68 +8,47 @@
 
 - [Charts](#charts)
 - [Supported Configurations](#supported-configurations)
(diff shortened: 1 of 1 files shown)


mgreau (approved):
LGTM 👍

PR #1361, opened 2021-09-21
[meta] remove contributing file from 7.15 branch

jmlrt:
This commit removes the CONTRIBUTING.md doc file from the 7.15 branch.
This file isn't used as the different charts README are pointing to the
file in master branch instead of 7.15 branch.
In addition this will simplify the release branch update workflow.


ygel (approved):
LGTM

Issue #1375, opened 2021-09-22
[elasticsearch] Readiness probe is failing with 8.0.0-SNAPSHOT and default config


jmlrt:
**Chart version:** `8.0.0-SNAPSHOT`

**Kubernetes version:** all

**Kubernetes provider:** all

**Helm Version:** all

**`helm get release` output**: 

<details>
<summary>Output of helm get release</summary>

```
NAME: helm-es-default
LAST DEPLOYED: Wed Sep 22 15:49:36 2021
NAMESPACE: default
STATUS: failed
REVISION: 1
USER-SUPPLIED VALUES:
null

COMPUTED VALUES:
antiAffinity: hard
antiAffinityTopologyKey: kubernetes.io/hostname
clusterHealthCheckParams: wait_for_status=green&timeout=1s
clusterName: elasticsearch
enableServiceLinks: true
envFrom: []
esConfig: {}
esJavaOpts: ""
esMajorVersion: ""
extraContainers: []
extraEnvs: []
extraInitContainers: []
extraVolumeMounts: []
extraVolumes: []
fsGroup: ""
fullnameOverride: ""
healthNameOverride: ""
hostAliases: []
httpPort: 9200
image: docker.elastic.co/elasticsearch/elasticsearch
imagePullPolicy: IfNotPresent
imagePullSecrets: []
imageTag: 8.0.0-SNAPSHOT
ingress:
  annotations: {}
  enabled: false
  hosts:
  - host: chart-example.local
    paths:
    - path: /
  tls: []
initResources: {}
keystore: []
labels: {}
lifecycle: {}
masterService: ""
maxUnavailable: 1
minimumMasterNodes: 2
nameOverride: ""
networkHost: 0.0.0.0
networkPolicy:
  http:
    enabled: false
  transport:
    enabled: false
nodeAffinity: {}
nodeGroup: master
nodeSelector: {}
persistence:
  annotations: {}
  enabled: true
  labels:
    enabled: false
podAnnotations: {}
podManagementPolicy: Parallel
podSecurityContext:
  fsGroup: 1000
  runAsUser: 1000
podSecurityPolicy:
  create: false
  name: ""
  spec:
    fsGroup:
      rule: RunAsAny
    privileged: true
    runAsUser:
      rule: RunAsAny
    seLinux:
      rule: RunAsAny
    supplementalGroups:
      rule: RunAsAny
    volumes:
    - secret
    - configMap
    - persistentVolumeClaim
    - emptyDir
[...]

PR #1384, opened 2021-10-06
[elasticsearch] use security by default

jmlrt:
This commit update Elasticsearch chart to use security by default.

- Adds a new Secret templates for Elasticsearch credentials with a
  randomized password if password value isn't defined.

- Adds instructions to retrieve credentials in Elasticsearch chart
  deployment notes.

- Update Kibana, Filebeat and Metricbeat charts to use Elasticsearch
  credentials from the new Elasticsearch chart secret by default.

Fix https://github.com/elastic/helm-charts/issues/1375


jmlrt:
Merging as Elasticsearch chart tests are successful and the other charts will be fixed in follow-up PRs.

PR #1419, opened 2021-10-15
Adds logstash UDP service 

### LINKED ISSUE #1041: Deploy 2 load balancers from the logstash chart
**Chart version:**
7.7.0

**Kubernetes version:**
1.8

EKS

**Helm Version:**
2.17

I'm using a nlb load balancer for TCP connections, but I would also need another load balancer for the UDP because it is not currently possible to mix TCP and UDP. Is it possible to deploy 2 load balancer from the logstash chart ? I was unsucessful adding another one after the first one, only the last one is deployed.

framsouza:
This PR enable to create logstash using UDP service option. 

Fix #1041


jmlrt (commented):
Thanks for this PR @framsouza,

Depending on the answer to https://github.com/elastic/helm-charts/issues/1041#issuecomment-965036878, we may not need it anymore.

framsouza:
@jmlrt as per my comment [here](**url**), I think we can close this one

PR #1422, opened 2021-10-18
[metricbeat] Adding custerRole permissions for leader election

### LINKED ISSUE #1415: [metricbeat] Missing clusterRole permissions for leader election in k8s 1.19+
Chart version:

7.15+

With the leader election enabled in the kubernetes autodiscovery and running on a cluster 1.19+ the leader election fails due to a lack of permissions in the clusterRole being used.

The below
https://github.com/elastic/helm-charts/blob/e077086c74d3b9e121cec1d93f04d8418599bae4/metricbeat/values.yaml#L240-L261

needs to be expanded with:
```yaml
  - apiGroups: ["coordination.k8s.io"]
    resources:
      - leases
    verbs: ["create", "get", "list", "update"]
```
Steps to reproduce:
1. Deploy on a 1.19 or later k8s cluster
2. Configure metricbeat with the following excerpt (add output destination as needed)
   ```yaml
      metricbeatConfig:
        metricbeat.yml: |
          # enable the http endpoint for health checks
          http:
            enabled: true
            host: localhost
            port: 5066

          # autodiscover kubernetes nodes and collect the metrics from the workers
          metricbeat.autodiscover:
            providers:
              # this uses leader election to have one run as the master and have
              # the leader scrape the kube-state-metrics endpoint and the
              # kubernetes API endpoint
              - type: kubernetes
                scope: cluster
                node: ${NODE_NAME}
                unique: true
                identifier: leader-election-metricbeat
                templates:
                  - config:
                      # kubernetes state metrics
                      - module: kubernetes
                        hosts: ["kube-state-metrics.metrics:8080"]
                        period: 10s
                        add_metadata: true
                        metricsets:
                          - state_node
                          - state_deployment
                          - state_daemonset
                          - state_replicaset
                          - state_pod
                          - state_container
                          - state_job
                          - state_cronjob
                          - state_resourcequota
                          - state_statefulset
                          - state_service
                      # API server metrics
                      - module: kubernetes
                        metricsets:
                          - apiserver
                        hosts: ["https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}"]
                        bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
                        ssl.certificate_authorities:
                          - /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
                        period: 30s
                      # kubernetes events
                      - module: kubernetes
                        metricsets:
                          - event
   ```

Without this the pod logs will be filled with:
```
error retrieving resource lock metrics/metricbeat-cluster-leader: leases.coordination.k8s.io "metricbeat-cluster-leader" is forbidden: User "system:serviceaccount:metrics:metricbeat-metricbeat" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "metrics"
```

framsouza:
This PR adds permissions to metricbeat use leader election (auto discovery) 

Fix #1415


jmlrt (approved):
LGTM⛴

Do you know if we may need this new rule for some other charts?

jmlrt (commented):
Looking at https://github.com/elastic/beats/blob/ac8275f72e55b699fb5ae41c68774280d7188bd3/deploy/kubernetes/metricbeat/metricbeat-role.yaml#L46-L59, we may need to add this to a simple role instead of a cluster role.

Seems related to https://github.com/elastic/beats/pull/24958/files

jmlrt:
> LGTM⛴
> 
> Do you know if we may need this new rule for some other charts?

We may also add it to Filebeat chart => https://github.com/elastic/beats/blob/ac8275f72e55b699fb5ae41c68774280d7188bd3/deploy/kubernetes/filebeat/filebeat-role.yaml#L22-L35

framsouza:
@jmlrt Just added a single role for filebeat and metricbeat, can you please have a look at it?

metricbeat/values.yaml
```diff
@@ -259,6 +259,10 @@ clusterRoleRules:
     resources:
       - nodes/stats
     verbs: ["get"]
+  - apiGroups: ["coordination.k8s.io"]
+    resources:
+      - leases
+    verbs: ["create", "get", "list", "update"]
```
jmlrt, line 265: I think this should also be added to https://github.com/elastic/helm-charts/blob/c52ad079379e6b1ba420866cdf96019f382f94ff/filebeat/values.yaml#L203-L213

jmlrt (changes requested):

metricbeat/values.yaml
bskim45: I guess just role binding is ok? https://github.com/elastic/beats/blob/878e778aed26ca30a660157aa2e11cd9f6a899ef/deploy/kubernetes/metricbeat-kubernetes.yaml#L315

PR #1423, opened 2021-10-20
Add support to use modules on filebeat 

### LINKED ISSUE #1036: Cannot enable Filebeat modules through values.yml
**Chart version:**
7.9.2

**Kubernetes version:**

EKS

**Helm Version:**
3.2.0

I try to enable modules from values file, but it didn't work.
values.yaml configuration:
```
...
  filebeatConfig:
    filebeat.yml: |
      filebeat.modules:
      - module: nginx
      - module: mysql
      - module: system
      filebeat.config:
        modules:
          enabled: true
          path: ${path.config}/modules.d/*.yml
...
```

When i do helm install with this config in values file, nothing happen, i tried to list enabled modules inside the container with filebeat modules list, and nginx, mysql, system are still disabled. Configuration is copied in the container's filebeat.yml file tho.  

I am able to enable them only with filebeat modules enable, but this method is not persistent. 

What can be the reason? I tried everything, moving filebeat.modules part after filebeat.config and so on, nothing help.

framsouza:
This PR adds support to mount modules.d on filebeat.

It will create a extra ConfigMap pointing to modules.d directory and a new entry on values.yml which will contain the modules configuration like:

```
  filebeatModules: {}
    system.yml: |
      - module: system
        syslog:
          enabled: true
```

It will give you the following output:
```
kubectl exec fb-filebeat-rnxpp -- filebeat modules list
Enabled:
system
```

Fix #1036


framsouza:
jenkins test this please

filebeat/README.md
```diff
@@ -124,6 +124,7 @@ as a reference. They are also used in the automated testing of this chart.
 | `deployment.extraVolumes`      | Templatable string of additional `volumes` to be passed to the `tpl` function for Deployment                                                                                 | `[]`                               |
 | `daemonset.hostAliases`        | Configurable [hostAliases][] for filebeat Deployment                                                                                                                         | `[]`                               |
 | `deployment.filebeatConfig`    | Allows you to add any config files in `/usr/share/filebeat` such as `filebeat.yml` for filebeat Deployment                                                                   | see [values.yaml][]                |
+| `deployment.filebeatModules`   | Allows you to add any module configuration in `/usr/share/filebeat/modules.d/` such as `system.yml` for filebeat system module                                               | see [values.yaml][]                |
```
jmlrt, line 127: There should also be a line for `daemonset.filebeatModules`

jmlrt (changes requested):
Not fully tested yet but the Filebeat change LGTM.
An additional line should be added in the README for `deployment.filebeatModules`.
In addition, we should have some Python tests for it.

The Metricbeat changes don't seem related.

PR #1440, opened 2021-11-02
Adding NODE_IP env to replace NODE_NAME.

kosalaat:
…(available from 1.7.0-alpha.1) instead of spec.nodeName (available from 1.4.0-alpha.3).

Some of the cloud enviroments (AKS, Azure RH OpenShift) seem to have trouble resolving node names. statu.hostIP would provide a IP instead of the name.


ebuildy:
hello @kosalaat , thanks you for the PR, 

It looks like this is already possibe via ``values.yaml`` , using ``extraEnvs`` field to add NODE_IP env, isnit?

Maybe this would be better to add a new example folder, with correct values, to support AKS/Azure/...?

kosalaat:
hi @ebuildy,

I'm happy to contribute back with what we learn in the field. I used extraVars as a workaround and overwrite with NODE_IP instead of NODE_NAME. I was suggesting we change to NODE_IP, instead of NODE_NAME, so that would work for all the platforms.. and one less name resolution in the process :).

I noticed that Metricbeat is anyway supported from  1.14 onwards, and the versions checks I put in are not required. I'll close this PR, and I can resubmit if you think changing over to NODE_IP is a better approach :).

I'm happy to add the same to examples in a separate PR.

Cheers,
Ko

Issue #1443, opened 2021-11-05
[elasticsearch] Readiness probe is failing again with 8.0.0-SNAPSHOT and default config


jmlrt:
**Chart version:** `8.0.0-SNAPSHOT`

**Kubernetes version:** all

**Kubernetes provider:** all

**Helm Version:** all

**`helm get release` output**: 

<details>
<summary>Output of helm get release</summary>

```
NAME: es
LAST DEPLOYED: Fri Nov  5 17:27:14 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
USER-SUPPLIED VALUES:
null

COMPUTED VALUES:
antiAffinity: hard
antiAffinityTopologyKey: kubernetes.io/hostname
clusterHealthCheckParams: wait_for_status=green&timeout=1s
clusterName: elasticsearch
enableServiceLinks: true
envFrom: []
esConfig: {}
esJavaOpts: ""
esMajorVersion: ""
extraContainers: []
extraEnvs: []
extraInitContainers: []
extraVolumeMounts: []
extraVolumes: []
fsGroup: ""
fullnameOverride: ""
healthNameOverride: ""
hostAliases: []
httpPort: 9200
image: docker.elastic.co/elasticsearch/elasticsearch
imagePullPolicy: IfNotPresent
imagePullSecrets: []
imageTag: 8.0.0-SNAPSHOT
ingress:
  annotations: {}
  className: nginx
  enabled: false
  hosts:
  - host: chart-example.local
    paths:
    - path: /
  pathtype: ImplementationSpecific
  tls: []
initResources: {}
keystore: []
labels: {}
lifecycle: {}
masterService: ""
maxUnavailable: 1
minimumMasterNodes: 2
nameOverride: ""
networkHost: 0.0.0.0
networkPolicy:
  http:
    enabled: false
  transport:
    enabled: false
nodeAffinity: {}
nodeGroup: master
nodeSelector: {}
persistence:
  annotations: {}
  enabled: true
  labels:
    enabled: false
podAnnotations: {}
podManagementPolicy: Parallel
podSecurityContext:
  fsGroup: 1000
  runAsUser: 1000
podSecurityPolicy:
  create: false
  name: ""
  spec:
    fsGroup:
      rule: RunAsAny
    privileged: true
    runAsUser:
      rule: RunAsAny
    seLinux:
      rule: RunAsAny
    supplementalGroups:
      rule: RunAsAny
    volumes:
    - secret
    - configMap
[...]

PR #1458, opened 2021-12-09
[elasticsearch] use bash for readiness script

jmlrt:
This commit set the readiness probe to use bash instead of sh. This is
required for Elasticsearch > 7.16.0 because the Docker image is now
based on Ubuntu instead of CentOS 8, and sh on Ubuntu isn't compatible
with the `if [[ ... -eq .... ]]` statements used in the readiness probe.

cc @elastic/es-delivery

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -227,7 +227,7 @@ spec:
         readinessProbe:
           exec:
             command:
-              - sh
+              - bash
               - -c
               - |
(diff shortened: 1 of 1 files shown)


jmlrt:
The tests aren't expected to work because the Elasticsearch chart on the main branch is still broken due to #1443. However, this should be backported to `7.x` branch to fix `7.x`  CI tests failing.

pugnascotia (approved):
LGTM 👍

mgreau (approved):
LGTM

PR #1459, opened 2021-12-09
[7.x] [elasticsearch] use bash for readiness script (#1458)

jmlrt:
Backports the following commits to 7.x:
 - [elasticsearch] use bash for readiness script (#1458)

PR #1460, opened 2021-12-09
[meta] download goss outside of pods

jmlrt:
This commit update the way goss binary is retrieved in the
tests/examples. The goss binary is now downloaded locally then copied
inside the pod.

This is needed because the Elasticsearch Ubuntu based Docker images is
missing the ca-certificates package, and so the curl https command is
failling with `curl: (77) error setting certificate verify locations`
error.

Follow-up of https://github.com/elastic/helm-charts/pull/1458
Relates to https://github.com/elastic/helm-charts/pull/1459#issuecomment-989987927

### DIFF
--- a/helpers/examples.mk
+++ b/helpers/examples.mk
@@ -13,6 +13,7 @@ help: ## Display this help
 goss: ## Run goss tests
 	set -e; \
 	for i in $$(seq 1 5); do \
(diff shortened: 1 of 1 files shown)


Conky5 (approved):
LGTM

PR #1469, opened 2021-12-13
[metricbeat] remove es metricset search query for oss example

jmlrt:
This commit removes the goss test in charge of querying Elasticsearch
search API for container metricset in Metricbeat oss example.

This is because the default Elasticsearch used in this test is also used
by the default Metricbeat and so the search API usually return indices
for default Metricbeat for this query in the test environment.


mgreau (approved):
LGTM

framsouza (approved):
LGTM

PR #1409, opened 2021-10-11
[all] Add support to ingress networking.k8s.io/v1 & ingressClassName

framsouza:
This PR adds support to ingress `networking.k8s.io/v1` that was added in[ Kubernetes v1.19](https://kubernetes.io/docs/concepts/services-networking/ingress/).

Fix #1381
Fix #1136

PR #1478, opened 2021-12-14
[7.16] [all] Add support to ingress networking.k8s.io/v1 & ingressClassName (#1409)

jmlrt:
Backports the following commits to 7.16:
 - [all] Add support to ingress networking.k8s.io/v1 & ingressClassName (#1409)


jmlrt:
transient error

PR #1301, opened 2021-07-20
[kibana] optionally disable SA token automount

jonkerj:
### Benefits
By disabling the automount, potential attackers cannot access the Kubernetes API on behalf/through the pod.

### Possible drawbacks
If anyone is using some sidecar or plugin to access the Kubernetes API, they will have to explicitly enable (`--set automountToken=true`) the automount of the SA token in the values.

### Applicable issues
#1330 

- [x] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [x] README.md updated with any new values or changes
- [x] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

PR #1482, opened 2021-12-14
[kibana] optionally disable SA token automount (#1301)

jmlrt:
Kibana has no direct interaction with the Kubernetes API, and as such,
it does not need a mounted service account token in its pods. By
disabling this automount, potential attackers cannot access the API on
behalf/through the Pod.

This commit allows users to opt out on SA token automount, but leaves
leaves current behaviour untouched to avoid breaking things.

Signed-off-by: Jorik Jonker <jorik.jonker@eu.equinix.com>


jmlrt:
jenkins test this please

jmlrt:
jenkins test this please

PR #1506, opened 2021-12-20
[elasticsearch] remove usage of ELASTIC_USERNAME

jmlrt:
This commit remove the usage of ELASTIC_USERNAME and replace it by
hardcoded "elastic" username.

This is required because ELASTIC_USERNAME don't seem to be supported
anymore supported by Elasticsearch (I couldn't find the exact version
where support was stopped).

This was already done as part of https://github.com/elastic/helm-charts/pull/1384 in the main branch.


jmlrt:
transient issues, merging

PR #1507, opened 2021-12-20
[elasticsearch] remove usage of ELASTIC_USERNAME (#1506)

jmlrt:
This commit remove the usage of ELASTIC_USERNAME and replace it by
hardcoded "elastic" username.

This is required because ELASTIC_USERNAME don't seem to be supported
anymore supported by Elasticsearch (I couldn't find the exact version
where support was stopped).

Backport of #1506


Conky5 (approved):
LGTM

PR #1530, opened 2022-01-05
[kibana] use bash for readiness script

### LINKED ISSUE #1529: [kibana] Readiness probe failed: `[[: not found` with 7.17.0-SNAPSHOT images
Kibana readiness probe is failling with `[[: not found` error when using Kibana chart from 7.x branch with 7.17.0-SNAPSHOT images.

This seem to be related to Kibana base image change from CentOS to Ubuntu in https://github.com/elastic/kibana/pull/119520. 

`[[ ... ]]` test syntax isn't available in the shell provided in Ubuntu base image:
https://github.com/elastic/helm-charts/blob/31ea1dd8376a8688e65ac739f8ff6c474db078ca/kibana/templates/deployment.yaml#L129

Similar to https://github.com/elastic/helm-charts/pull/1458 for Elasticsearch.

cc @elastic/kibana-operations

jmlrt:
This commit set the readiness probe to use bash instead of sh. This is
required for Kibana > 7.17.0 because the Docker image is now
based on Ubuntu instead of CentOS 8, and sh on Ubuntu isn't compatible
with the if [[ ... -eq .... ]] statements used in the readiness probe.

Fix #1529

### DIFF
--- a/kibana/templates/deployment.yaml
+++ b/kibana/templates/deployment.yaml
@@ -108,7 +108,7 @@ spec:
 {{ toYaml .Values.readinessProbe | indent 10 }}
           exec:
             command:
-              - sh
+              - bash
               - -c
               - |
                 #!/usr/bin/env bash -e


Conky5 (approved):
LGTM

PR #1531, opened 2022-01-05
[7.x] [kibana] use bash for readiness script (#1530)

jmlrt:
 - [kibana] use bash for readiness script (#1530)


jmlrt:
jenkins test this please

PR #1537, opened 2022-01-09
Fix the servicePort missing

### LINKED ISSUE #1500: [logstash 7.16] Can't set custom ingress port anymore in logstash chart
**Chart version:**
7.16.1

**Describe the bug:**

We use different TCP ports for different inputs on logstash. We expose only handful of them via external ingress.

After 7.16 upgrade, possibility to set custom servicePort for ingress has been removed from logstash chart for unknown reason.

Now, it is automatically set equal to httpPort value in configuration, which does not allow us to expose custom port on ingress anymore and instead forces us to use httpPort setting.
https://github.com/elastic/helm-charts/blob/b4d75e5c0902c44350ab9c7c779c0932b84bf66b/logstash/templates/ingress.yaml#L49

The change was introduced with this commit:
https://github.com/elastic/helm-charts/commit/1459b2f28e80d2abf33443568092feb6dac33328
       

In other charts, however, it is still possible to use custom servicePort, for example elasticsearch 7.16 chart:
https://github.com/elastic/helm-charts/blob/b4d75e5c0902c44350ab9c7c779c0932b84bf66b/elasticsearch/templates/ingress.yaml#L60

Is there any reason for restricting it for logstash?

Thank you.

HenItzhaki:

Fix #1500

### DIFF
--- a/logstash/templates/ingress.yaml
+++ b/logstash/templates/ingress.yaml
@@ -46,7 +46,7 @@ spec:
           service:
             name: {{ $fullName }}
             port:
-              number: {{ $httpPort }}
+              number: {{ .servicePort | default $httpPort  }}
     {{- end }}
   {{- end }}
 {{- end }}


Ricardolaponder:
Hi, I'm also hitting this issue, will you still merge this?

jmlrt:
cla/check

jmlrt:
Replaced by #1569

PR #1567, opened 2022-02-07
fix: headless service as ingress backend

Dirc:
- [.] README.md updated with any new values or changes
  n.a.
- [.] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`
  n.a. (doesn't exist for ingress)
  
Hi all,

I wanted to expose the httpPort of the headless service via ingress, but ran into a bug.

For the ingress backend, the port and service do not match:

- Ingress backend port is the httpPort (which is also the port of the headless service)
- Ingress backend service is "logstash.fullname" (which equal the name of the ClusterIP service)

In order to expose the httpPort 9600, I needed to change the ingress backend service into:
"logstash.fullname"-headless

Cheers,
Dirc

### DIFF
--- a/logstash/templates/ingress.yaml
+++ b/logstash/templates/ingress.yaml
@@ -44,7 +44,7 @@ spec:
         pathType: {{ $pathtype }}
         backend:
           service:
-            name: {{ $fullName }}
+            name: {{ $fullName }}-headless
             port:
               number: {{ $httpPort }}
     {{- end }}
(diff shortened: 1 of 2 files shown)


Dirc:
This probably conflicts with:
#1500 

I will join the discussion there, since I believe both a ingress for the webapi port and the http plugin port are valuable.

jmlrt:
Hi @Dirc, headless service purpose is only for maintaining the network identity of the pods in the statefulset and shouldn't be exposed.
If you need to expose a service, you should create another one using the `service` value instead of trying to reuse the headless one.

Dirc:
Hi @jmlrt , thanks for explaining. However, I still wonder, why is the httpPort (normally 9600) then the default port for the ingress backend service?

PR #1569, opened 2022-02-09
[logstash] add flexible ingress

### LINKED ISSUE #1500: [logstash 7.16] Can't set custom ingress port anymore in logstash chart

jmlrt:
Update Logstash ingress to be consistent with Elasticsearch and Kibana
ingresses:
- add back custom ingress port support
- allow to use single ingress path
- also improve example values and tests

Fix #1500


mgreau (dismissed):
LGTM

jmlrt:
👋🏻 @framsouza can you take a look 🙏🏻

framsouza (dismissed):

PR #1570, opened 2022-02-09
[logstash] add externalTrafficPolicy support

### LINKED ISSUE #1548: Support externalTrafficPolicy in the Logstash Helm chart
**Describe the feature:**

Setting the externalTrafficPolicy to Local does not appear to have any effect in the resulting service.

```
  service:
    type: LoadBalancer
    externalTrafficPolicy: Local
```

```
kubectl describe svc logstash-logstash

...(omitted)
External Traffic Policy:  Cluster
...
```

There is a number of reasons why someone would need to set the externalTrafficPolicy to Local, namely:

There are two available options: Cluster (default) and Local. Cluster obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading. Local preserves the client source IP and avoids a second hop for LoadBalancer and NodePort type Services, but risks potentially imbalanced traffic spreading.

from https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

jmlrt:
Fix #1548


framsouza (approved):
LGTM

PR #1573, opened 2022-02-14
(ElasticSearch) Add override for initial_master_nodes and seed_hosts

### LINKED ISSUE #1268: support multiple master node groups
**Describe the feature:**

As a Helm Charts user I would like to have an option to specify multiple master node groups with different settings. Today, `cluster.initial_master_nodes` is set to a list of nodes from current node group instead of _all_ master nodes which is what is advised in [ES documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.13/modules-discovery-bootstrap-cluster.html)

> You must set `cluster.initial_master_nodes` to the same list of nodes on each node on which it is set in order to be sure that only a single cluster forms during bootstrapping and therefore to avoid the risk of data loss.

Let's assume a setup with 3 master/data nodes in 3 different AZs.

```
AZ1 :
master/data node (node attribute az:az1)

AZ2:
master/data node (node attribute az:az2)

AZ3:
master/data node (node attribute az:az3)
```

It's preferred to configure shard allocation awareness in such scenario:
```
cluster.routing.allocation.awareness.attributes: az
cluster.routing.allocation.awareness.force.zone.values: az1,az2,az3
```

To achieve this multiple node groups must be used but they do not work correctly with master node groups.

PPACI:

Hi Elasticsearch.
This short PR introduce an override of template values.
This would close #1268 by enabling full override of `cluster.initial_master_nodes`.

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -312,16 +312,16 @@ spec:
                 fieldPath: metadata.name
           {{- if has "master" .Values.roles }}
           - name: cluster.initial_master_nodes
-            value: "{{ template "elasticsearch.endpoints" . }}"
+            value: "{{- tpl .Values.initial_master_nodes . -}}"
           {{- end }}
           - name: node.roles
             value: "{{ template "elasticsearch.roles" . }}"
           {{- if lt (int (include "elasticsearch.esMajorVersion" .)) 7 }}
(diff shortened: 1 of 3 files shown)


PPACI:
I've now signed the Contributor Agreement

ebuildy:
Hello, nice one!

Could you:
* use camelCase to name values
* add some tests

thanks you

Issue #1455, opened 2021-12-07
statefulset does not point to my custom service account name


mdnfiras:
**Chart version:**

version: 7.15.0

**Kubernetes version:**

v1.22.4

**Kubernetes provider:**

kubeadm

**Helm Version:**

using fluxcd `v0.24.0`, which has helm-controller `v0.14.0`, which implements helm `v3.6.3`

**Describe the bug:**

when setting `rbac.create` to `true` and `rbac.serviceAccountName` to a custom service account name like `elasticsearch-logging-sa`, the service account will be created with the custom name but the statefulset will keep pointing to the default service account name.

**Steps to reproduce:**

1. chose set values in the `values.yaml` file or using `--set`:

```
rbac:
  create: true
  serviceAccountName: "elasticsearch-logging-sa"
```

2. install the helm release

**Expected behavior:**

a service account with name `elasticsearch-logging-sa` is created and the statefulset points to that name.

**Provide logs and/or server output (if relevant):**

```bash
/ # kubectl get sa -n logging
NAME                       SECRETS   AGE
default                    1         145m
elasticsearch-logging-sa   1         143m
```

```bash
/ # kubectl describe sts elasticsearch-logging-master -n logging | tail -4
Events:
  Type     Reason        Age                 From                    Message
  ----     ------        ----                ----                    -------
  Warning  FailedCreate  80s (x18 over 12m)  statefulset-controller  create Pod elasticsearch-logging-master-0 in StatefulSet elasticsearch-logging-master failed error: pods "elasticsearch-logging-master-0" is forbidden: error looking up service account logging/elasticsearch-logging-master: serviceaccount "elasticsearch-logging-master" not found
```

mdnfiras:
can this pull request solve this problem? https://github.com/elastic/helm-charts/pull/1456/files

ebuildy:
Hello, I think this normal if you want the Helm release create it for you. All ressources generated by the Helm release must have the same naming convention (ConfigMap, Sts, SA etc...).

mdnfiras:
but the helm chart currently creates a service account named `rbac.serviceAccountName` if its provided, and `rbac.create` is set to `true`.

if all resources of the helm release must follow the chart's naming convention, then the helm chart should ignore the `rbac.serviceAccountName` field and create the service account with the default name when `rbac.create` is set to `true`.

ebuildy:
[...]

PR #1581, opened 2022-02-23
[logstash] fix ServiceAccount inconsistencies

jmlrt:
This commit refactor the way we define the ServiceAccount name to fix an
issue where a ServiceAccount is created with a custom name but the
Statefulset is trying to use a different ServiceAccount.

Relates to #1455 and 1580


framsouza (approved):
LGTM

PR #1417, opened 2021-10-14
Support to add Kibana annotations at deployment level

framsouza:
This PR adds support to insert Kibana annotations at Deployment level

PR #1599, opened 2022-03-03
[7.17] Support to add Kibana annotations at deployment level (#1417)

jmlrt:
 - Support to add Kibana annotations at deployment level (#1417)


jmlrt:
ok to test

PR #1604, opened 2022-03-04
[meta] update docker images

jmlrt:
- use python slim base images
- bump python version
- bump python dependencies
- bump tools version


mgreau (approved):
LGTM

PR #1605, opened 2022-03-07
Remove replicas field from deployment when autoscaling is enabled

mcwienczek:
ARGOCD requires to omit replicas in deployment when using HPA autoscaling, otherwise Argocd app will always be out-of-sync. See here: https://argo-cd.readthedocs.io/en/stable/user-guide/best_practices/#leaving-room-for-imperativeness


mcwienczek:
I have signed the CLA, but somehow github bot still says I haven't.

PR #1606, opened 2022-03-07
[7.17] [filebeat] add missing rolebing and cluster role rules (#1602)

jmlrt:
 - [filebeat] add missing rolebing and cluster role rules (#1602)


jmlrt:
ok to test

PR #1524, opened 2021-12-27
[metricbeat] bump kube-state-metrics to version 4.7.0

beatkind:
I have bumped the kube-state-metrics version referenced in the metricbeat chart. Otherwise, metricbeat does not work for k8s higher than 1.21.X.

- [X] Chart version *not* bumped (the versions are all bumped and released at the same time)
- [ ] README.md updated with any new values or changes
- [ ] Updated template tests in `${CHART}/tests/*.py` 
- [ ] Updated integration tests in `${CHART}/examples/*/test/goss.yaml`

Fix #1576 
Replace #1577

PR #1612, opened 2022-03-09
[metricbeat] fix upgrade test

jmlrt:
This commit fix the Metricbeat upgrade test by bumping the source
version to 7.17.1. This is required because
https://github.com/elastic/helm-charts/pull/1524 introduced a new
kube-state-metrics version that prevent upgrade from previous versions.


framsouza (commented):
LGTM

PR #1614, opened 2022-03-10
add ingress option to redirect to https

anubisg1:
when elasticsearch runs in secure mode, the internal service runs on https.
without nginx.ingress.kubernetes.io/backend-protocol: "HTTPS" , ingress would forward requests to http://service:9200 instead of the correct https://service:9200

- [x ] Chart version *not* bumped (the versions are all bumped and released at the same time)


anubisg1:
signed :)

PR #1623, opened 2022-03-17
[logstash] using https to connect to elasticsearch

framsouza:
This PR will use https to connect into elasticsearch by default


logstash/values.yaml
```diff
@@ -15,7 +15,7 @@ logstashConfig: {}
 
 # Allows you to add any pipeline files in /usr/share/logstash/pipeline/
 ### ***warn*** there is a hardcoded logstash.conf in the image, override it first
-logstashPipeline: {}
+logstashPipeline:  {}
```
jmlrt, line 18: NIT
```suggestion
logstashPipeline: {}
```

logstash/values.yaml
```diff
@@ -62,7 +62,7 @@ secrets: []
 #      cert.key.filepath: "secrets.crt" # The path to file should be relative to the `values.yaml` file.
 
 # A list of secrets and their paths to mount inside the pod
-secretMounts: []
+secretMounts: []
```
jmlrt, line 65: NIT
```suggestion
secretMounts: []
```

jmlrt (commented):
- Some spaces added by mistakes to remove
- Let's verify that the next CI test is OK for logstash

The rest LGTM.

PR #1630, opened 2022-03-23
[apm-server] using https to connect to elasticsearch

framsouza:
This PR enable apm-server to connect into Elasticsearch via https


framsouza:
ok to test

PR #1660, opened 2022-05-24
make ELASTIC_PASSWORD env var optional

acjohnson:

With the following values I am able to deploy elasticsearch 8.1.0 with security disabled, with one exception.
```
--set createCert=false \
--set protocol="http" \
--set secret.enabled=false \
--set extraEnvs[0].name="xpack.security.enabled" \
--set-string extraEnvs[0].value="false" \
```
The `readinessProbe` has an explicit requirement for `ELASTIC_PASSWORD` being set which needs to be optional in this case. 

Hoping I'm going down the right road here, maybe there is something I missed but if not it would be nice if you could merge :smiley:

### DIFF
--- a/elasticsearch/templates/statefulset.yaml
+++ b/elasticsearch/templates/statefulset.yaml
@@ -241,12 +241,6 @@ spec:
               - |
                 set -e
 
-                # Exit if ELASTIC_PASSWORD in unset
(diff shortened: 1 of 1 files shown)


acjohnson:
I completed the contributor agreement if you can run that check again...

PR #1676, opened 2022-08-02
[elasticsearch] fix helm test

jmlrt:
This commit fix the Elasticsearch Helm test by adding security
compatibility.

PR #1677, opened 2022-08-02
[elasticsearch] do not regenerate certs during upgrade

jmlrt:
This commit remove the `pre-upgrade` hook for the certificate secret to
ensure that the certificate isn't regenerate each time we run the
`helm upgrade` command.

### DIFF
--- a/elasticsearch/templates/secret-cert.yaml
+++ b/elasticsearch/templates/secret-cert.yaml
@@ -10,7 +10,7 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
   annotations:
-    "helm.sh/hook": "pre-install,pre-upgrade"
+    "helm.sh/hook": "pre-install"
     "helm.sh/hook-delete-policy": "before-hook-creation"
 data:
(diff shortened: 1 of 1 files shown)


jmlrt:
👋🏻 @framsouza, I discovered this issue while working on https://github.com/elastic/helm-charts/pull/1676. Each time I ran `helm upgrade`, the certificate was regenerated and the secret updated without updating the Elasticsearch pods, then running `helm test` was failing because it was using the new wrong certificate. 

I'm afraid we could also have issues in prod environnements when an Elasticsearch pod is recreated (for example because of a K8S node failure) in a chart release that has already been upgraded, because the new pod could be using a new certificate while the other ones would still be using the old certificate from the first deployment.

Do you see some use case where regenerating certs during upgrade is required?

jmlrt:
ok to test

jmlrt:
ok to test

mgreau (approved):
LGTM

framsouza (approved):

framsouza:
LGTM

PR #1690, opened 2022-09-07
[meta] set jenkins workers to ubuntu

jmlrt:
This commit is updating the helm-charts Jenkins jobs configuration to
use Ubuntu nodes only. This change is mostly to fix a bug with the ampersands
(`&`) in the node labels.

The only real change here is replacing every occurrence of `docker&&virtual`
by `ubuntu`. All other changes are simply yaml formatting done by VS Code.


jonahbull (approved):
LGTM

PR #1519, opened 2021-12-22
[elasticsearch] SSL by default

framsouza:
Fix #1443 

This PR add ssl configuration by default in our helm chart. The SSL certs are generated by helm.

PR #1691, opened 2022-09-07
[elasticsearch] do not regenerate certs when they already exists

jmlrt:
This commit update the way we generate the Elasticsearch certs to lookup
if the certs secret already exists and reuse the secrets instead of regenerate
the certs.

This fix the issue when upgrading from a previous version of the chart
that doesn't include the certs secret by ensuring that the certs secret
is created if it doesn't exists.

This method also add more flexibility than charts hooks by ensuring that
the certs won't be rotated each time we upgrade the chart.

Found in https://stackoverflow.com/a/61715870

### DIFF
--- a/elasticsearch/templates/_helpers.tpl
+++ b/elasticsearch/templates/_helpers.tpl
@@ -28,15 +28,22 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 
 {{/*
-Generate certificates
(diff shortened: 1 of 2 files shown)


jmlrt:
cc @framsouza for review as this is impacting your changes from https://github.com/elastic/helm-charts/pull/1519/. Not sure if there is another way to handle it.

framsouza (approved):

PR #1696, opened 2022-09-13
[filebeat][metricbeat] increase memory limit for goss tests

jmlrt:
This commit increase the memory limit for Filebeat and Metricbeat goss
tests. This should help with the regular `Error 137` during goss tests.

Indeed `Error 137` seems to be related to OOM. As beats pods default
memory limit is really small (200Mb), it seems that injecting goss
inside the pods to run the test can reach the memory limit regularly.

Also remove the biggest test in the default config test examples to make
goss use less memory because we can't override default values for these
tests.


jmlrt:
All Filebeat and Metricbeat tests are ✅

framsouza (approved):

PR #1701, opened 2022-09-14
[7.17] [filebeat][metricbeat] increase memory limit for goss tests (#1696)

jmlrt:
 - [filebeat][metricbeat] increase memory limit for goss tests (#1696)


jmlrt:
ok to test

Issue #1714, opened 2022-10-27
[kibana] initContainer: configure-kibana-token: Back-off restarting failed container: 


mdnfiras:
**Chart version:**
8.4.1 (from the main branch in [this point in history](https://github.com/elastic/helm-charts/tree/8565b30cde0c373bbd2812079c2114bfcf8ed524))

**Kubernetes version:**
1.21

**Kubernetes provider:**
GKE (Google Kubernetes Engine)

**Describe the bug:**
Kibana's `initContainer` `configure-kibana-token` keep crashing forever.

**Steps to reproduce:**

1. Deploy elasticsearch 8 helm chart and enable security features (username/password + SSL)
2. Deploy kibana 8 helm chart and reference the appropriate elasticsearch credentials and certificates secrets
3. After kibana's pod successfully runs for the first time, delete it
4. the new kibana pod will have its `initContainer` `configure-kibana-token` crashing forever.

**Expected behavior:**
the new kibana pod will have its `initContainer` `configure-kibana-token` completes successfully.

**Provide logs and/or server output (if relevant):**
`configure-kibana-token` `initContainer` logs before crashing:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 409 Conflict
```
this init container creates a token for kibana's service account and saves it for kibana's actual container.

if i run a similar command from within the elasticsearch pods: `curl -k -u $ELASTIC_USERNAME:$ELASTIC_PASSWORD -XPOST https://localhost:9200/_security/service/elastic/kibana/credential/token/mykibana8-kibana?pretty` I get the following response:
```json
{
  "error" : {
    "root_cause" : [
      {
        "type" : "version_conflict_engine_exception",
        "reason" : "[service_account_token-elastic/kibana/mykibana8-kibana]: version conflict, document already exists (current version [1])",
        "index_uuid" : "zwuGuJcSS2OUs1ClsFdThB",
        "shard" : "0",
        "index" : ".security-7"
      }
    ],
    "type" : "version_conflict_engine_exception",
    "reason" : "[service_account_token-elastic/kibana/mykibana8-kibana]: version conflict, document already exists (current version [1])",
[...]

PR #1715, opened 2022-10-27
Fix kibana's init container configure-kibana-token http request error 409

mdnfiras:
this is a fix proposal to this issue: https://github.com/elastic/helm-charts/issues/1714

kubernetes pods die regularly and new ones are created. kibana's init container does not work this way since it always tries to create a new token, which elasticsearch will deny if it already exists.

this PR makes kibana's init container attempt to delete the old token before attempting to create a new one. deleting a non-existing token returns a valid status code so this is not a breaking change. I chose this approach because I found no way of getting an existing token using elasticsearch API, the token is only returned on creation (please correct me if i'm wrong).

Also, i noticed that there is a helm cleanup job to cleanup the old token, but that one runs only when we delete the entire helm release, and will not trigger on pod termination/creation.

### DIFF
--- a/kibana/templates/deployment.yaml
+++ b/kibana/templates/deployment.yaml
@@ -93,7 +93,9 @@ spec:
         command:
           - sh
           - -c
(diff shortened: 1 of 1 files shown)


jmlrt:
Closing because it will have to be done in a different way in https://github.com/elastic/helm-charts/pull/1720.

Thanks for your work @mdnfiras

PR #1679, opened 2022-08-04
[kibana] add 8.x compatibility

jmlrt:
This PR is updating the Kibana chart to make it compatible with 8.x version.

Issue #603, opened 2020-04-30
Use multiple elasticsearch hosts


refucktor:
**Chart version:**
7.6.2

**Kubernetes provider:** E.g. GKE (Google Kubernetes Engine)
AWS

**Helm Version:**
2.16.6

**Describe the bug:**
Trying to set multiple values in the `elasticsearchHosts`, turns out that is not able to parse the values. I already try to set the value in multiple ways. Could somebody just put a WORKING example of it.
 Here are some of the formats that I tried:

```
## Array example 1:
kibana.yml:
  elasticsearchHosts: '["http://es1:9200", "http://es2:9200", "http://es3:9200"]'

## Array example 2:
kibana.yml:
  elasticsearchHosts: ["http://es1:9200", "http://es2:9200", "http://es3:9200"]

## Array example 3:
kibana.yml:
  elasticsearchHosts:
  - http://es1:9200
  - http://es2:9200
  - http://es3:9200

## Array example 4:
kibana.yml:
  elasticsearchHosts:
    http://es1:9200
    http://es2:9200
    http://es3:9200
```

**Expected behavior:**
The helm is able to parse multiple values as array

**Provide logs and/or server output (if relevant):**
```
 FATAL  Error: [config validation of [elasticsearch].hosts]: types that failed validation:
- [config validation of [elasticsearch].hosts.0]: expected URI with scheme [http|https] but got [[http://es1:9200].
- [config validation of [elasticsearch].hosts.1]: expected value of type [array] but got [string]

```

discostur:
having exactly the same problem

abuisine:
The solution would be something like this : 
```
elasticsearchHosts: "[\\\"http://192.168.69.61:9200\\\",\\\"http://192.168.69.62:9200\\\",\\\"http://192.168.69.63:9200\\\"]"
```
I am about to propose a PR as it is insane :)

Madmaxguy:
Still an issue and also experiencing it

torgus:
Hi, Any news on this? No way to have an HA Kibana if we can not configure all ES nodes.

sk1me:
Any news on that feature being implemented? 🥲

PR #1720, opened 2022-11-04
[kibana] create a secret with es token

jmlrt:
This PR moves the Elasticsearch token for Kibana creation into Helm pre-install hooks and adds it to a K8S secret so it can be used by every pod of the Kibana Deployment.

The K8S secret and all linked resources are cleaned in post-delete hooks.

The call to Elasticsearch and K8S API to create the ES token and convert it to a secret are done in a Node JS script because the only language runtime available in the Kibana Docker image are Node and Bash, but `jq` command isn't available which exclude Bash script.

The Node JS script is also available in plain JS format to facilitate the review: https://gist.github.com/jmlrt/482b3d984588c1d3e10a84ce5fcd7b3b

Many thanks @pugnascotia for the help on this script 🙏🏻 

Follow-up of https://github.com/elastic/helm-charts/pull/1679
Replace https://github.com/elastic/helm-charts/pull/1715
Fixes https://github.com/elastic/helm-charts/issues/1714

### DIFF
--- a/kibana/templates/configmap-helm-scripts.yaml
+++ b/kibana/templates/configmap-helm-scripts.yaml
@@ -171,4 +171,5 @@ data:
         break;
       default:
         console.log('Unknown command');
+        process.exit(1);
     }
--- a/kibana/templates/configmap-helm-scripts.yaml
+++ b/kibana/templates/configmap-helm-scripts.yaml
@@ -17,10 +17,8 @@ data:
     const fs = require('fs');
 
     // Elasticsearch API
-    //TODO: remove hardcoded host
-    const esHostname = "elasticsearch-master";
-    const esPort = process.env.ELASTICSEARCH_MASTER_SERVICE_PORT;
     const esPath = '_security/service/elastic/kibana/credential/token/kb-kibana';
+    const esUrl = '{{ .Values.elasticsearchHosts }}' + '/' + esPath
     const esUsername = process.env.ELASTICSEARCH_USERNAME;
     const esPassword = process.env.ELASTICSEARCH_PASSWORD;
     const esAuth = esUsername + ':' + esPassword;
@@ -31,16 +29,17 @@ data:
     const k8sHostname = process.env.KUBERNETES_SERVICE_HOST;
     const k8sPort = process.env.KUBERNETES_SERVICE_PORT_HTTPS;
     const k8sPath = 'api/v1/namespaces/{{ .Release.Namespace }}/secrets';
+    const k8sUrl = 'https://' + k8sHostname + ':' + k8sPort + '/' + k8sPath
     const k8sBearer = fs.readFileSync('/run/secrets/kubernetes.io/serviceaccount/token');
     const k8sCa = fs.readFileSync('/run/secrets/kubernetes.io/serviceaccount/ca.crt');
 
     // With thanks to https://stackoverflow.com/questions/57332374/how-to-chain-http-request
-    function requestPromise(options, payload) {
+    function requestPromise(url, options, payload) {
       return new Promise((resolve, reject) => {
-        const request = https.request(options, response => {
+        const request = https.request(url, options, response => {
 
           console.log('statusCode:', response.statusCode);
-          console.log('headers:', response.headers);
+          // console.log('headers:', response.headers);
 
           // TODO: remove 404 and handle it during esToken deletion
           const isSuccess = response.statusCode >= 200 && response.statusCode < 300 || response.statusCode == 404;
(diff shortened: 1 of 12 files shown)


jmlrt:
👋🏻 @jbudz @pugnascotia the Node JS script is working fine but could surely be optimized, advises are welcome.

I'm also looking for a better way to handle the case where the Elasticsearch token doesn't exist yet by not having to consider that HTTP 404 is successful (TLDR, when trying to delete an ES token that doesn't exist, it's returning 404) if you have some better idea.

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,161 @@ kind: ConfigMap
...
-    set -euo pipefail
+  manage-es-token.js: |
+    const https = require('https');
+    const fs = require('fs');
 
-    echo "export ELASTICSEARCH_SERVICEACCOUNTTOKEN=$({{ template "kibana.home_dir" . }}/node/bin/node {{ template "kibana.home_dir" . }}/helm-scripts/parse-token.js {{ template "kibana.home_dir" . }}/config/tokens/{{ template "kibana.fullname" . }}.json)" > $HOME/.elasticsearch-serviceaccounttoken
-    source $HOME/.elasticsearch-serviceaccounttoken
+    // Elasticsearch API
+    const esPath = '_security/service/elastic/kibana/credential/token/{{ template "kibana.fullname" . }}';
+    const esUrl = '{{ .Values.elasticsearchHosts }}' + '/' + esPath
```
pugnascotia, line 22: Will `.Values.elasticsearchHosts` always only have a single value?

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,161 @@ kind: ConfigMap
...
+    const esUsername = process.env.ELASTICSEARCH_USERNAME;
+    const esPassword = process.env.ELASTICSEARCH_PASSWORD;
+    const esAuth = esUsername + ':' + esPassword;
```
pugnascotia, line 25: Should we validate that the username and password have been supplied? I would be better to catch that explicitly if we'd misconfigured or failed to propagate something.

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,161 @@ kind: ConfigMap
...
 
-    # https://github.com/elastic/dockerfiles/blob/a405a4d692031b72cefcb8523bd464aa3221ec47/kibana/Dockerfile#L131
-    exec /bin/tini -- /usr/local/bin/kibana-docker "$@"
+    // Kubernetes API
+    const k8sHostname = process.env.KUBERNETES_SERVICE_HOST;
+    const k8sPort = process.env.KUBERNETES_SERVICE_PORT_HTTPS;
+    const k8sPostSecretPath = 'api/v1/namespaces/{{ .Release.Namespace }}/secrets';
+    const k8sDeleteSecretPath = 'api/v1/namespaces/{{ .Release.Namespace }}/secrets/{{ template "kibana.fullname" . }}-es-token';
+    const k8sPostSecretUrl = 'https://' + k8sHostname + ':' + k8sPort + '/' + k8sPostSecretPath;
+    const k8sDeleteSecretUrl = 'https://' + k8sHostname + ':' + k8sPort + '/' + k8sDeleteSecretPath;
```
pugnascotia, line 35: Optional:

```suggestion
    const k8sPostSecretUrl = `https://${k8sHostname}:${k8sPort}/${k8sPostSecretPath}`;
    const k8sDeleteSecretUrl = `https://${k8sHostname}:${k8sPort}/${k8sDeleteSecretPath}`;
```

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,161 @@ kind: ConfigMap
...
+        'Content-Type': 'application/json',
+      }
+    };
+
+    // With thanks to https://stackoverflow.com/questions/57332374/how-to-chain-http-request
+    function requestPromise(url, options, payload) {
+      return new Promise((resolve, reject) => {
+        const request = https.request(url, options, response => {
+
+          console.log('statusCode:', response.statusCode);
```
pugnascotia, line 74: Do we need to log this?

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,161 @@ kind: ConfigMap
...
+      }).catch(err => {
+        console.error(err);
+      });
+    }
+
+    const command = process.argv[2];
+    switch (command) {
+      case 'create':
+        console.log('Creating a new Elasticsearch token for Kibana')
+        createEsToken();
```
pugnascotia, line 156: I think it would be cleaner to always return the promises created by `requestPromise`, so that they are chained together, and put a single top-level `.catch(...)` here that logs any errors. If that isn't clear, I can put the suggested changes in a rustpad.

kibana/templates/configmap-helm-scripts.yaml
jmlrt: Good question, it seems it was supposed to support multiple values but a bug that was never fixed prevented it (https://github.com/elastic/helm-charts/issues/603). 

As this Kibana chart is supposed to connect to an Elasticsearch cluster deployed via the Elastic Helm charts, all Elasticsearch pods are supposed to be available via a single K8S service that should be used in `.Values.elasticsearchHosts`.

IMHO, It's not worth fixing https://github.com/elastic/helm-charts/issues/603 and making the script compatible with multiple URLs. I'll update the doc to mention that a single URL (Elasticsearch service URL) is expected for this value, to make it clear.

kibana/templates/configmap-helm-scripts.yaml
jmlrt: `ELASTICSEARCH_USERNAME` and `ELASTICSEARCH_PASSWORD` variables are defined in the Job pod template where this script is supposed to run:

https://github.com/elastic/helm-charts/blob/81ce165618d2b7602bff2e814e9825800c087aa0/kibana/templates/pre-install-job.yaml#L28-L37

If these variables can't be filled (because the K8S secret doesn't exist or doesn't have the expected format), the job will not start and not trigger the script. So theoretically, there is no need to add validation, but I agree, this would be better 👍🏻

I'll try to provide an update for that.

kibana/templates/configmap-helm-scripts.yaml
jmlrt: This is not required but is still helpful to troubleshoot without users having to modify the script in case something is wrong.

kibana/templates/configmap-helm-scripts.yaml
jmlrt: ff4a1c60b02b3553120dc97454519a1917e90085

kibana/templates/configmap-helm-scripts.yaml
jmlrt: Do you mean something like that?
```javascript
function createEsToken() {
  // Chaining requests
  console.log('Cleaning previous token');
  return requestPromise(esUrl, esTokenDeleteOptions).then(() => {
    console.log('Creating new token');
    requestPromise(esUrl, esTokenCreateOptions).then(response => {
      // ...
      requestPromise(k8sPostSecretUrl, secretCreateOptions, secretData)
    });
  });
}

function cleanEsToken() {
  // Chaining requests
  console.log('Cleaning token');
  return requestPromise(esUrl, esTokenDeleteOptions).then(() => {
    // Create the k8s secret
    console.log('Delete K8S secret');
    requestPromise(k8sDeleteSecretUrl, secretDeleteOptions)
  });
}

const command = process.argv[2];
switch (command) {
  case 'create':
    console.log('Creating a new Elasticsearch token for Kibana')
    createEsToken().catch(err => {
      console.error(err);
    });
    break;
  case 'clean':
    console.log('Cleaning the Kibana Elasticsearch token')
    cleanEsToken().catch(err => {
      console.error(err);
    });
    break;
  default:
    console.log('Unknown command');
}

```

pugnascotia: Yes, just make sure you're always returning the value of `requestPromise`, and the chain of promises will finally resolve in the `switch`.

jmlrt:
> I'm also looking for a better way to handle the case where the Elasticsearch token doesn't exist yet by not having to consider that HTTP 404 is successful (TLDR, when trying to delete an ES token that doesn't exist, it's returning 404) if you have some better idea.

improved in 43c08a13100b29023ec7640094d6623532aa067e

kibana/templates/configmap-helm-scripts.yaml
jmlrt: done in e3ef0c1620c5953b623d232babf63489b67e9bb5

jmlrt:
@pugnascotia PTAL 🙏🏻 
I also updated the plain JS version in https://gist.github.com/jmlrt/482b3d984588c1d3e10a84ce5fcd7b3b (and updated the issue description with the good link^^)

jmlrt:
> > 👋🏻 @jbudz @pugnascotia the Node JS script is working fine but could surely be optimized, advises are welcome.
> > I'm also looking for a better way to handle the case where the Elasticsearch token doesn't exist yet by not having to consider that HTTP 404 is successful (TLDR, when trying to delete an ES token that doesn't exist, it's returning 404) if you have some better idea.
> 
> I could reproduce it by setting Elasticsearch replicas to 0 and running the script manually:
> 
> ```
> kibana@pre-install-kb-kibana-wh4lm:~$ node/bin/node helm-scripts/manage-es-token.js create
> Creating a new Elasticsearch token for Kibana
> Cleaning previous token
> DELETE undefined failed:  connect ECONNREFUSED 10.44.6.135:9200
> Error: connect ECONNREFUSED 10.44.6.135:9200
>     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1187:16) {
>   errno: -111,
>   code: 'ECONNREFUSED',
>   syscall: 'connect',
>   address: '10.44.6.135',
>   port: 9200
> }
> kibana@pre-install-kb-kibana-wh4lm:~$ echo $?
> 0
> ```
> 
> Despite logging the `ECONNREFUSED` error, `manage-es-token.js` still returns success (RC 0).
> 
> Need to ensure that the script returns a different RC when it is failing with this error.

should be fixed by 86386f0

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,172 @@ kind: ConfigMap
...
 
-    echo "export ELASTICSEARCH_SERVICEACCOUNTTOKEN=$({{ template "kibana.home_dir" . }}/node/bin/node {{ template "kibana.home_dir" . }}/helm-scripts/parse-token.js {{ template "kibana.home_dir" . }}/config/tokens/{{ template "kibana.fullname" . }}.json)" > $HOME/.elasticsearch-serviceaccounttoken
-    source $HOME/.elasticsearch-serviceaccounttoken
+    // Read environment variables
+    function getEnvVar(name) {
+      if (name in process.env) {
+        return process.env[name]
+      } else {
+        throw new Error(name + ' environment variable is missing')
+      }
```
pugnascotia, line 26: This would evaluate to true if a value existed in the environment but was an empty string. I'd probably just go with the following, and let JS do it's "is it truthy" thing with empty vs non-empty strings.

```suggestion
      if (!process.env[name]) {
        throw new Error(name + ' environment variable is missing')
      }
      return process.env[name]
```

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,172 @@ kind: ConfigMap
...
+        break;
+      case 'clean':
+        console.log('Cleaning the Kibana Elasticsearch token')
+        cleanEsToken().catch(err => {
+          console.error(err);
+          process.exit(1);
+        });
+        break;
+      default:
+        console.log('Unknown command');
```
pugnascotia, line 174: We should probably exit non-zero in this case too:

```suggestion
        console.log('Unknown command');
        process.exit(1);
```

pugnascotia (approved):

jmlrt:
ok to test

jmlrt:
https://devops-ci.elastic.co/job/elastic+helm-charts+pull-request+integration-filebeat/1343/FILEBEAT_SUITE=default,KUBERNETES_VERSION=1.24,label=ubuntu/ is a known transient issue, just waiting for a few more review to merge.

mgreau (approved):
LGTM 👍

kibana/templates/configmap-helm-scripts.yaml
```diff
@@ -4,18 +4,172 @@ kind: ConfigMap
...
+        console.log('Creating new token');
+        requestPromise(esUrl, esTokenCreateOptions).then(response => {
+          const body = JSON.parse(response);
+          const token = body.token.value
+
+          // Encode the token in base64
+          const base64Token = Buffer.from(token, 'utf8').toString('base64');
+
+          // Prepare the k8s secret
+          secretData = JSON.stringify({
```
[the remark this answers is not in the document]
jbudz: whoops I missed the merge, didn't refresh in time.  Sorry about the confusion.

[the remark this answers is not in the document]
jbudz: (I'll follow up with a PR)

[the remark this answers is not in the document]
jbudz: https://github.com/elastic/helm-charts/pull/1727

PR #1727, opened 2022-11-15
[kibana/manage-es-token] scope/promise updates

jbudz:
This makes a few minor changes:
- return nested promises to preserve the promise chain
- scope secretData to the current block


jmlrt (approved):
LGTM⛴

